PR #3, opened 2015-05-02
markdown renderer cleanups; fix links with stylized content

vito:
This should fix things like:

```markdown
[`foo bar`](baz)
```

Did some refactoring along the way to switch things to a more recursive style, which lends itself more to crazy things like `**_this_**`.

### DIFF
--- a/scribble-lib/scribble/markdown-render.rkt
+++ b/scribble-lib/scribble/markdown-render.rkt
@@ -6,8 +6,6 @@
          scribble/text/wrap)
 (provide render-mixin)
 
-(define current-preserve-spaces (make-parameter #f))
-
 (define current-indent (make-parameter 0))
 (define (make-indent amt)
   (+ amt (current-indent)))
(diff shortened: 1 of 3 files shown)


mflatt:
@greghendershott Can you take a look at these changes?

greghendershott:
Disclaimer: It feels like a _very_ long time since I did this. I do remember trying to hew closely to the plain-text renderer because I felt a bit lost in the Scribble rendering model and flow.

In other words I have more confidence in @vito 's ability to grasp the code than my own.

I did read through the changes and they look good to me. And I see there is a new test case for the new feature, which is good.

mflatt:
Merged - thanks!
I had to adjust the "markdown.md" test output to restore some trailing whitespace, since that whitespace is generated by the markdown renderer. I assume that the whitespace removal wasn't intentional (i.e., wasn't intended to check that the renderer avoids some whitespace).

vito:
Yeah. My editor probably just stripped it out.

Thanks!

PR #4, opened 2015-05-08
Two bugfixes.

elibarzilay:
* Fix contract on `with-writer`.

* The default writer is initially `write-string`, not `display`.


mflatt:
merged

PR #8, opened 2015-06-28
convert blueboxes.rkt to typed racket

AlexKnauth:


scribble-lib/scribble/blueboxes.rkt
```diff
@@ -44,14 +65,19 @@
         (class/interface-tag->constructor-tag
          (definition-tag->class/interface-tag tag))
         cache))
-     (append plain-strs
-             (if constructor-strs '("") '())
-             (if constructor-strs (cdr constructor-strs) '()))]
+     (assert
```
rfindler: This assert is because TR can't track the non-emptyness across the append call, right?

scribble-lib/scribble/blueboxes.rkt
```diff
@@ -135,8 +175,10 @@
                (error 'build-blueboxes-cache
                       "blueboxes info didn't have the right shape: ~s"
                       candidate))
-             candidate))
-         (set-bluebox-info-offset! a-bluebox-info (and desed (+ (string->number first-line) pos)))
+             (cast candidate Blueboxes-Info-Hash)))
+         (when (and desed (string? first-line))
+           (define first-line-num (assert (string->number first-line) exact-nonnegative-integer?))
```
rfindler: I think that probably this assert can actually fail (if the file is corrupt or serialization changes somehow or something) and so it should go thru the same path as line 169 (in the new code) where we log a warning.

rfindler:
Thanks for doing this! I put some comments inline.

scribble-lib/scribble/blueboxes.rkt
AlexKnauth: Yes.

scribble-lib/scribble/blueboxes.rkt
AlexKnauth: Oh, I missed that one.

AlexKnauth: I fixed it

AlexKnauth: And there was another one that I missed too, that I also fixed.

scribble-lib/scribble/blueboxes.rkt
```diff
@@ -115,17 +155,19 @@
...
      (call-with-input-file blueboxes.rktd
-       (λ (port)
+       (λ ([port : Input-Port])
          (port-count-lines! port)
          (define first-line (read-line port))
          (define pos (file-position port))
          (define desed 
-           (with-handlers ([exn:fail? (λ (x) 
+           (with-handlers ([exn:fail? (λ (x)
+                                        (assert x exn:fail?)
```
samth: You should be able to avoid this `assert` by annotating the binding of `x`.

scribble-lib/scribble/blueboxes.rkt
samth: I think we should fix this by just improving the type for `append` in Typed Racket.

scribble-lib/scribble/blueboxes.rkt
AlexKnauth: Ok, I hadn't realized that. Thanks!

rfindler:
I'm finding this hard to follow on github. But it looks like there is still an assert on line 102 that should probably not be a runtime error, but is instead somethign that drracket should swallow and continue as if there was no blueboxes information. Is that right?

scribble-lib/scribble/blueboxes.rkt
```diff
@@ -62,30 +88,42 @@
...
+                   (file-position port (+ (car offset+len) (or offset 0)))
                    (for/list ([i (in-range (cdr offset+len))])
+                     : (Listof (U String EOF))
                      (read-line port)))))))
           (cond
             [(ormap eof-object? lines) #f]
-            [else lines])]
-         [else #f])])))
+            [(null? lines) #f]
+            [else (assert (filter string? lines) pair?)])]
```
samth: Is this the `assert` you mean, @rfindler?

rfindler: Yes.

AlexKnauth: I put the `filter string?` there because the type-checker didn't seem to notice the `(ormap elf-object? lines)` before that, so the `filter string?` should just act like identity.  Then I put the `(null? lines)` clause in to catch empty lists and return `#f` instead, and then the `(assert ... pair?)` because it doesn't know that `filter string?` should act like identity. That's probably not the best way of expressing those intentions clearly, but that was how that got there.

samth: This seems like it changes the behavior from the original program on `null` -- previously the `ormap` produced `#f` and so just `lines` is returned, but now `#f` is returned.

Once you drop that `cond` clause, I'd just write `(cast line (Pair String (Listof String)))` as a clearer statement of intent.

`ormap` could have a better negative filter to improve type checking here, and probably remove the need for casts at all.

AlexKnauth: Yes, but since the `fetch-strs-for-single-tag` function is supposed to return either a non-empty list or `#f`, the previous behavior was a bit wrong (according to the contract on `fetch-blueboxes-strs` that was there before).

rfindler: On Sun, Jun 28, 2015 at 4:20 PM, Alex Knauth <notifications@github.com>
wrote:

> In scribble-lib/scribble/blueboxes.rkt
> <https://github.com/racket/scribble/pull/8#discussion_r33427136>:
>
> >                       (read-line port)))))))
> >            (cond
> >              [(ormap eof-object? lines) #f]
> > -            [else lines])]
> > -         [else #f])])))
> > +            [(null? lines) #f]
> > +            [else (assert (filter string? lines) pair?)])]
>
> Yes, but since the fetch-strs-for-single-tag function is supposed to
> return either a non-empty list or #f, the previous behavior was a bit
> wrong (according to the contract on fetch-blueboxes-strs that was there
> before).
>
>

Cool!! :)

AlexKnauth:
Somehow (with this pull request) it got into a state where it's giving me DrRacket Internal Errors like this every time I try to type something:
```
application: procedure does not expect an argument with given keyword
  procedure: fetch-blueboxes-strs
  given keyword: #:blueboxes-dirs
  arguments...:
   #t
   #:blueboxes-dirs '(#<path:/Applications/Racket/2015-06-08/Racket v6.2.0.4/doc/afl> #<path:/Applications/Racket/2015-06-08/Racket v6.2.0.4/doc/alexis-util> #<path:/Applications/Racket/2015-06-08/Racket v6.2.0.4/doc/algol60> #<path:/Applications/Racket/2015-06-08/Racket v...
  context...:
   /Applications/Racket/2015-06-08/Racket v6.2.0.4/collects/racket/private/kw.rkt:1296:14
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/racket/drracket/drracket/drracket/private/syncheck/blueboxes-gui.rkt:671:2: fetch
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/racket/drracket/drracket/drracket/private/syncheck/blueboxes-gui.rkt:670:0: get-blueboxes-cache
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/racket/drracket/drracket/drracket/private/syncheck/blueboxes-gui.rkt:444:4: compute-tag+rng method in ...ck/blueboxes-gui.rkt:173:2
   /Applications/Racket/2015-06-08/Racket v6.2.0.4/share/pkgs/gui-lib/framework/private/coroutine.rkt:47:20
```
But what's weird is that if I look at `drracket/private/syncheck/blueboxes-gui.rkt:671:2: fetch`, it's a call to `(make-blueboxes-cache #t #:blueboxes-dirs (get-rendered-doc-directories #f #f))` and not a call to `fetch-blueboxes-strs`.  Is there any possible way that it could be mixing them up or something, or is there another explanation, or?

rfindler:
I am not seeing that. I cherry picked the commit
f1be5573b4a315fd33adc7e0dde488c4f08e4cc1 into my scribble checkout and then
ran 'raco setup -D' and things seem to be working normally.

(I do see that some dependencies need to be added.)

Robby

AlexKnauth:
Running `raco setup` seemed to fix the problem for me. And I see the missing deps, I'll add them.

AlexKnauth:
I took away the `assert` on the `append`, since append now does preserve the non-emptyness of the first argument.

rfindler:
....  how do you feel about putting the close parens back so they aren't on their own lines?

And other than that, is there any reason not to push this to the repo?

AlexKnauth:
My experience with close-parens on the same line as the last item of the list is that if it's something like a deps specification, a require form, or a provide form, then it creates extra noise, especially with git diffs, when I want to add an item to the end of the list. 

If I understand correctly, that's why http://docs.racket-lang.org/style/Textual_Matters.html#%28part._.Where_to_.Put_.Parentheses%29 says 
> You are allowed to place all closing parenthesis on a line by itself at the end of long sequences, be those definitions or pieces of data.
> Doing so is most useful when you expect to add, delete, or swap items in such sequences.

rfindler:
Ah, okay. I guess I'm too set in my ways; I think that the diff algorithms
should adjust!

Okay with you to push now?

Robby

AlexKnauth:
Yes

AlexKnauth:
Or not:
I just got this:
```
application: procedure does not accept keyword arguments
  procedure: blueboxes-cache?
  arguments...:
   'dotmethod
   #:blueboxes-cache #<blueboxes-cache>
  context...:
   /Applications/Racket/2015-06-08/Racket v6.2.0.4/collects/racket/private/kw.rkt:1296:14
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/racket/drracket/drracket/drracket/private/syncheck/blueboxes-gui.rkt:444:4: compute-tag+rng method in ...ck/blueboxes-gui.rkt:173:2
   /Applications/Racket/2015-06-08/Racket v6.2.0.4/share/pkgs/gui-lib/framework/private/coroutine.rkt:47:20
```

rfindler:
Also, I think the unstable-contract-lib dependency is not necessary.

As for the error, are all .zos rebuilt? (re-run raco setup with no args)
I'm not getting an error and I think I got the latest commit.

Robby

AlexKnauth:
`raco setup` will probably fix it, but if it breaks this easily because of some bug somewhere else, maybe it's not a good idea? This definitely shouldn't be happening, I don't think it's because of anything in this pull request, but is it possibly a bug in typed racket somewhere, or?

I didn't get this error right away either, and I'm not sure what triggered it, but it's happened to me twice now.

rfindler:
I think that if you use non-(up to date) .zo files, then arbitrary badness
can ensue and that's not TR's fault. You really should never be loading a
.zo file that was compiled against an old version of some library. One
could, I think, reasonably say that this is a problem with the racket
runtime system, but if I understand correctly, it is judged too expensive
to fix. In the past the only issue that you were likely to get was a
specific error which everyone learned meant "recompile, please" so it
wasn't a big deal either. But that's now different with keyword
optimizations and cross-module inlining and perhaps other things too. It
would be nice if we could do some checking and signal an error (or
automatically recompile, as appropriate), but that seems beyond the purview
of this PR.

How about you make sure your .zos are up to date and if you don't see the
error after some reasonable amount of time doing other things with
DrRacket, then I'll push the commit? And if it keeps coming back evne with
up to date .zo files then we'll investigate further?

AlexKnauth:
Ok It's been a while and I haven't had this problem again, so I think it's good.

rfindler:
Okay, I've pushed the commit.

Thanks!

PR #10, opened 2015-07-07
Fix eval:check in interaction to properly validate its arguments

### LINKED ISSUE #9: Using eval:check in interaction doesn't seem to work
Even this simple usage of `eval:check` fails:

```racket
#lang racket/base
(require scribble/eval)
(interaction (eval:check #t #t))
```

```
eval: example result check failed in: #t
```

What's wrong?

lexi-lambda:
Fixes #9.

### DIFF
--- a/scribble-lib/scribble/eval.rkt
+++ b/scribble-lib/scribble/eval.rkt
@@ -258,12 +258,12 @@
                              (get-outputs)))])
       (cons (render-value (do-plain-eval ev s #t)) (get-outputs))))
   (define (do-ev/expect s expect)
-    (define r (do-ev s))
     (when expect
-      (let ([expect (do-plain-eval ev (car expect) #t)])
-        (unless (equal? (car r) expect)
+      (let ([r (do-plain-eval ev s #t)]
+            [expect (do-plain-eval ev (car expect) #t)])
+        (unless (equal? r expect)
           (raise-syntax-error 'eval "example result check failed" s))))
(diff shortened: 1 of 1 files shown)


lexi-lambda:
@rfindler Quick ping to check if someone saw this and/or remembers it still exists. :wink: I'd appreciate if you (or someone else) could take a look, but there's no rush, obviously.

rfindler:
I didn't write eval:check, but from looking at the diff, I wonder if this is backwards compatible. Like it is going to require an extra quote now for some results?

lexi-lambda:
It was my understanding that `eval:check` is currently just broken—that is, it doesn't work at all. Perhaps I'm mistaken, but I couldn't figure out how to make it work properly.

lexi-lambda:
I did a little bit of digging to see where this code came from. I looks like `eval:check` was added in 2008 by @mflatt, and the relevant code appears to have been most recently modified by @elibarzilay. Can anyone confirm how recently this worked?

rfindler:
I cannot but I guess git is more reliable than anyone's memory anyways. I
would also look for test cases and uses of it in the docs too, just to be
safe.

Robby

lexi-lambda:
I didn't find anything at all grepping through the codebase for `eval:check` besides the implementation and the brief note in the docs. As far as I can tell, it's unused.

stamourv:
Merged, thanks!

PR #17, opened 2015-09-06
link the #lang in codeblocks

AlexKnauth:
This makes the codeblock form output a (hash-lang) element for the
#lang, which is consistent with racketblock


mflatt:
merged - thanks!

PR #20, opened 2015-09-16
Added example of @tech with #:doc

wilbowma:
I never remember how to use this.


wilbowma:
Also, it appears my editor stripped out trailing whitespace (which I added for Racket and Scribble files per http://docs.racket-lang.org/style/Textual_Matters.html#(part._.Spaces))

stamourv:
I agree that trailing whitespace should be avoided. However in this case, I think having the semantic change stand out is more valuable than cleaning up whitespace. I'll adjust the commit and merge.

stamourv:
Merged, thanks!

PR #26, opened 2016-01-07
allow for replacing more scribble tex files

florence:
Allow for replacing of `scribble.tex`.

With this instead of:
```
(latex-defaults
                        (string->bytes/utf-8 <your header>)
                        "style.tex"
                        (list <your.cls>))
```

one could now write:

```
(latex-defaults+replacements
                        (string->bytes/utf-8 <your header>)
                        "style.tex"
                        (list <your.cls>)
                        (hash "scribble.tex" "new-scribble-tex"))
```

to replace the `scribble.tex` with `new-scribble.tex`.

`latex-defaults+replacements` uses a hash map to allow for further extensions without needing to create a new struct.


mflatt:
This is ok with me.
I think we'd ideally have something more fine-grained, since any document that relies on a "scribble.tex" replacement is likely to fail if the source "scribble.tex" has to change. If package conflicts are the main source of trouble, would it make sense to break the package-loading part of "scribble.tex" into a separate file (that can be replaced with this new technique), so that just the packages can be adjusted? Or maybe each package load should be guarded with some macro, so that a "scribble-prefix.tex" replacement can suppress or modify individual package loads?

florence:
Given that `\renewcommand` would let someone override any part of this except for the package loading, i think breaking the package loading portion out into its own file and allowing that file to be replaced  works. I'll go set that up.

florence:
Based on Matthews and Matthias's comments I've updated this to allow for overriding individual package includes.

florence:
Merged at  2881ef2

PR #29, opened 2016-02-16
Add /Library/TeX/texbin as another candidate-dir

sorawee:
MacTeX, TeX for Mac, will install the binaries at `/Library/TeX/texbin`. Hence, we should add this path to the candidate-dirs.

### DIFF
--- a/scribble-lib/scribble/private/run-pdflatex.rkt
+++ b/scribble-lib/scribble/private/run-pdflatex.rkt
@@ -92,4 +92,4 @@
 ;; here so that the "scribble pdf" button is more 
 ;; likely to work in drracket
 (define macosx-candidate-dirs
-  '("/usr/texbin"))
+  '("/usr/texbin" "/Library/TeX/texbin"))


rfindler:
Are you seeing this not work? `which` reports that pdflatex is in `/Library/TeX/texbin/pdflatex` in my shell and the "scribble pdf" button works for me, even with a DrRacket started via double clicking in the finder.

sorawee:
Yup, it's not working for me...

soegaard:
Hi Sorawee,

I am in the process of testing this (due to the bug report on
racket-poppler)
Infact I am downloading TeXLive 2015 as I write this.
I notice that El Capitan has a "rootless" feature, which has affected how
TexLive installs TeX and friends.

Just checking: Is it TeXLive 2015 you are using?

/Jens Axel




2016-02-16 17:54 GMT+01:00 sorawee <notifications@github.com>:

> Yup, it's not working for me...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/racket/scribble/pull/29#issuecomment-184767283>.
>



-- 
-- 
Jens Axel Søgaard

sorawee:
I downloaded it from http://tug.org/mactex/mactex-download.html. Not sure if this is they are the same or not.

soegaard:
That's the right one.

Robby: This document  http://tug.org/mactex/UpdatingForElCapitan.pdf
describes the changes. What used to be /usr/texbin now needs
to be /Library/TeX/texbin.

I can't remember: Does Scribble already handle this?

/Jens Axel




2016-02-16 18:24 GMT+01:00 sorawee <notifications@github.com>:

> I downloaded it from http://tug.org/mactex/mactex-download.html. Not sure
> if this is they are the same or not.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/racket/scribble/pull/29#issuecomment-184785685>.
>



-- 
-- 
Jens Axel Søgaard

soegaard:
The new location used by TeX Live is not in the path - unless the user actively adds it himself.
Therefore it makes sense to add it. I tested the change without changing my path - and it works as expected.

PR #32, opened 2016-03-08
scribble/jfp: affiliation-mark example

bennn:
Small example using `author/short` with affiliations.

- Is there a better way to write this? I don't like the space before `"and"` on [line 63](https://github.com/racket/scribble/compare/master...bennn:affiliation-example?expand=1#diff-1c5234a2c4169b9cf9aec5e47350c27eR63)
- Is there a way to render this in the reference manual? `doc-render-examples` is not quite right; should I add a screenshot?


bennn:
Added a `#lang` to make the example self-contained. I'm happy with this. (@stamourv can you take a look?)

stamourv:
LGTM

bennn:
cool, thanks
https://github.com/racket/scribble/commit/7836b788095b3de152fe3b3aa95879feabbba8a4

PR #33, opened 2016-03-10
Make the list of special characters user-extensible .

stamourv:


dvanhorn:
Thanks @stamourv!   This is very close to what I'd like, but I think I would prefer to have a parameter that's a `char -> string or #f` function.  It's easy to implement such a function with a dictionary so I don't think anything is lost.  But it also let's me write `string` as a conversion function that works well with `xelatex`, i.e. I really want those characters to pass through to the tex file.

stamourv:
Merged a version that uses functions, as you suggested.

PR #39, opened 2016-04-16
Add svg tag and all html 5 tags (except: output, time, and, map)

soegaard:
This adds all tags in html 5 not already present in scribble/html.
To prevent identifier clashes the tags  output, time, and, map were omitted.

### DIFF
--- a/scribble-html-lib/scribble/html/html.rkt
+++ b/scribble-html-lib/scribble/html/html.rkt
@@ -163,12 +163,50 @@
   tr ; holds th or td
   th ; header cell
   td ; table cell
+  ;; =========== HTML 5
+  rb          ; Ruby base
+  nav         ; Section with navigational links
+  meter       ; Gauge
+  data        ; Machine-readable equivalent
+  footer      ; Footer for a page or section
+  figure      ; Figure with optional caption
+  template    ; Template  
+  figcaption  ; Caption for figure
+  progress    ; Progress bar
+  datalist    ; Container for options for combo box control
+  bdi         ; Text directionality isolation
+  audio       ; Audio player
+  aside       ; Sidebar for tangentially related content
+  article     ; Contact information for a page or article element
+  ruby        ; Ruby annotation(s)
+  main        ;
+  rt          ; Ruby annotation text
+  header      ; Introductory or navigational aids for a page or section
+  canvas      ; Scriptable bitmap canvas
+  mark        ; Highlight
+  video       ; Video player
+  rtc         ; Ruby annotation text container 
+  rp          ; Parenthesis for ruby annotation text
+  section     ; Generic document or application section
+  ;; =========== HTML 5 (omitted)
+  ; output      ; Calculated output value
+  ; time        ; Machine-readable equivalent of date- or time-related data
+  ; map         ; Image map
+  ;;=========== Other
+  svg         ; Scalable Vector Graphics
   )
 
 ;; [*] empty elements, these are listed with an `EMPTY' content in
 ;; http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd
 (define/provide-elements/empty
-  base meta link hr br basefont param img area input isindex col)
+  base meta link hr br basefont param img area input isindex col
+  ;; ==== html 5
+  keygen ; Cryptographic key-pair generator form control
+  track  ; Timed text track
+  wbr    ; Line breaking opportunity
+  source ; Media source for video or audio
+  embed ; Plugin
+  )
 
 ;; [*] elements with a cdata/comment body
 (provide script/inline)


elibarzilay:
1. The new tags are listed in what looks to me like random order?  -- It's not alphabetical and not logical (eg, the ruby stuff is scattered all over).

2. Adding new names might break existing code -- in particular a binding for `nav` sounds like it might break the web-generating code.  (There's probably a bunch of `nav` bindings there.)

3. Since this tends to be messy, I'd prefer it if there was instead a separate file for newer variants -- something like `scribble/html/5` that re-provides the old stuff except for ancient "about-time" deprecated things, and adds the new things.

4. And if you go for that, then maybe also add a similar `.../4` with the esoteric names in html4, so the library by default binds only a small subset that works in both variants, and the .../4 can also include the more esoteric things like `map`.

5. And speaking about such separation, maybe drop the ruby stuff (or leave just `ruby` in, making it similar to `svg`) into its own file and add a file for svg things?  It's especially bad that the ruby thing has 4 additional bindings, and they're all short as if to make it more likely to have namespacing issues...

soegaard:
2016-05-01 21:03 GMT+02:00 Eli Barzilay <notifications@github.com>:

>
>    1.
>
>    The new tags are listed in what looks to me like random order? -- It's
>    not alphabetical and not logical (eg, the ruby stuff is scattered all over).
>
>
The original tags are in the order they appear in the reference document.
I followed that so the new tags also appear in the order they appear in:

    https://www.w3.org/TR/html5/index.html#elements-1

I agree that alphabetical order for all tags would be better.

>
>    1.
>
>    Adding new names might break existing code -- in particular a binding
>    for nav sounds like it might break the web-generating code. (There's
>    probably a bunch of nav bindings there.)
>
> True - that's why I inquired here:

    https://groups.google.com/d/msg/racket-users/wCN3TVJzNzk/HU75uE66AgAJ

>
>    1.
>
>    Since this tends to be messy, I'd prefer it if there was instead a
>    separate file for newer variants -- something like scribble/html/5
>    that re-provides the old stuff except for ancient "about-time" deprecated
>    things, and adds the new things.
>
> Maybe scribble/html5  ?

I believe the existing tags matches xhtml.

>
>    1.
>
>    And if you go for that, then maybe also add a similar .../4 with the
>    esoteric names in html4, so the library by default binds only a small
>    subset that works in both variants, and the .../4 can also include the more
>    esoteric things like map.
>
> That makes sense.

>
>    1.
>
>    And speaking about such separation, maybe drop the ruby stuff (or
>    leave just ruby in, making it similar to svg) into its own file and
>    add a file for svg things? It's especially bad that the ruby thing has 4
>    additional bindings, and they're all short as if to make it more likely to
>    have namespacing issues...
>
> I was about to leave <ruby> out but looked it up. The tag is used for
showing prononunciation
of East Asian characters.

Truth be told - I don't need scribble/html anymore.

The problem is that the representation it produces is too tied up with
Scribble
to be helpful when working with html (I gave up writing a converter to
sexp).
But since scribble/html was designed to work with Scribble only, this is
what to be expected.

Instead I have written a small package with the same surface syntax which
produces
sexps directly.

(If you wonder about the numbering: Gmail is a little too smart and decided
to renumber
each segment from scratch - sigh)

/Jens Axel

elibarzilay:
Re ruby: Yeah, I saw what the ruby thing is, but it seems like a heavy hit in terms of name grabbing compared to relatively low use, which is why it sounds to me fine to require a library for the extra bindings.

As for your use: it's not at all tied up with Scribble (the documentation system) -- it's not even used by it IIRC.  When I designed it I went in a parallel tangent, with the goal of making websites (ie, the main one) and Matthew ended up in a very different direction for the documentation system (and he used xexprs, I think).  I think that there's value in using it for whatever you're doing...

soegaard:
2016-05-02 19:49 GMT+02:00 Eli Barzilay <notifications@github.com>:

> Re ruby: Yeah, I saw what the ruby thing is, but it seems like a heavy hit
> in terms of name grabbing compared to relatively low use, which is why it
> sounds to me fine to require a library for the extra bindings.
>
> As for your use: it's not at all tied up with Scribble (the documentation
> system) -- it's not even used by it IIRC. When I designed it I went in a
> parallel tangent, with the goal of making websites (ie, the main one) and
> Matthew ended up in a very different direction for the documentation system
> (and he used xexprs, I think). I think that there's value in using it for
> whatever you're doing...
>
After looking at scribble/html/xml.rkt again, I see that my problem was the
representation of literals.

https://github.com/racket/scribble/blob/master/scribble-html-lib/scribble/html/xml.rkt#L126

A literal is represented as a writer (from scribble/text). That
representation works fine
if printing is the only operation on literals, but it is a bit cumbersome
for other purposes
(such as converting to sexp).

/Jens Axel

elibarzilay:
Oh -- that's actually one of the limitations of the xexpr thing, IIRC -- that you cannot just dump a bunch of text to be placed in the output as-is...

bennn:
Closed by #59 , these tags are provided by the `scribble/html/extra` module.

PR #40, opened 2016-04-17
Provide the special struct

soegaard:
In order to write a converter from html expression represented using `element` and `special` into SXML the struct `special` needs to be provided.

### DIFF
--- a/scribble-text-lib/scribble/text/output.rkt
+++ b/scribble-text-lib/scribble/text/output.rkt
@@ -250,6 +250,7 @@
 
 ;; special constructs
 
+(provide (struct-out special))
 (define-struct special (flag contents))
 
 (define-syntax define/provide-special


elibarzilay:
I'm not sure about this one.  On one hand, I don't see why you need it --- adding more information at that level is running anyway into the problem of thunks that you need to run to get their text, so it seems to me like you want to work at the level of generate the text and use that--?

(This is unrelated to exposing the xml structs, which is something that I definitely wanted to do but didn't get to.  Perhaps something that is relevant here too is some function that takes in text with xml structs and "materializes" all the text except for the xml structs themselves?)

soegaard:
You are right. I thought that I would get just enough information such that
a converter into sexp
were possible. But as you say, it isn't enough.

/Jens Axel


2016-05-01 21:22 GMT+02:00 Eli Barzilay <notifications@github.com>:

> I'm not sure about this one. On one hand, I don't see why you need it ---
> adding more information at that level is running anyway into the problem of
> thunks that you need to run to get their text, so it seems to me like you
> want to work at the level of generate the text and use that--?
>
> (This is unrelated to exposing the xml structs, which is something that I
> definitely wanted to do but didn't get to. Perhaps something that is
> relevant here too is some function that takes in text with xml structs and
> "materializes" all the text except for the xml structs themselves?)
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/racket/scribble/pull/40#issuecomment-216065719>
>



-- 
-- 
Jens Axel Søgaard

Issue #25, opened 2015-12-31
in `scribble/lp2`, `chunk` should use `#:tag-prefix`


mbutterick:
Consider a project that gathers multiple `scribble/lp2` files into a single document. For TOC / navigation purposes, the names of the `chunk` blocks are treated like section titles. But unlike `section`, `subsection`, et al., they don't obey the `#:tag-prefix` setting, which means that to make things index correctly, you have to manually assign unique names to every `chunk` in the project.

Issue #15, opened 2015-08-05
Can't require a submodule in the same chunk with scribble/lp2


jsmaniac:
In the code below, the `(require 'm)` fails. Interestingly, it is possible to require `m` from a sibbling submodule works, but `(require (submod "." m))` doesn't work.

```
#lang scribble/lp2

@chunk[<*>
       (module m racket
         (define x 1)
         (provide x))
       (require 'm)                 ;; Fails with "require: unknown module" module name: #<resolved-module-path:'m>
       (require (submod "." m))     ;; Fails with "require: unknown module" module name: #<resolved-module-path:(submod 'scribble-lp-tmp-name m)>
       (module other racket
         (require (submod ".." m))) ;; Works.
       ]
```

jsmaniac:
Wrapping the module `m` and the `require` with `(begin (module m ...) (require 'm))` works fine. It feels like the names were not introduced in the same context, due to an overly-aggressive `strip-context` or `replace-context` in scribble/lp/lang/common.rkt, but I'm not sure.

```scheme
#lang scribble/lp2

@chunk[<*>
       (begin
         (module m racket
           (define x 1)
           (provide x))
         (require 'm))] ;; Works.
```

PR #43, opened 2016-06-17
Fixes scribble bug #25 (in `scribble/lp2`, `chunk` should use `#:tag-prefix`)

jsmaniac:
This will likely conflict with PR #15, if you merge #15 first, I'll update this so that it applies without conflicts.


jsmaniac:
This fixes issue #25 .

jsmaniac:
My bad, it would seem just using 'prefixable is not enough. I must have read the docs too quickly.

jsmaniac:
It seems that 'prefixable works only if we put it in the first item of what elemref calls a "tag", but which seems to be rather a taglet. Putting 'prefixable in the car of the second element of the list seems to not work.

PR #45, opened 2016-06-25
Add language-index-desc struct, subtype of module-path-index-desc

bennn:
Used to tell if a module implements a `#lang` or reader.
Required by https://github.com/racket/racket/pull/1355


mflatt:
Looks fine to me.

PR #48, opened 2016-07-07
Add reader-index-desc struct

bennn:
Thanks to @AlexKnauth for pointing out that `#reader` modules should not appear if a user searches for `#lang` modules.

This PR adds a new index description for reader modules. Also changes where `manual-mod.rkt` branches on its `lang` argument, to keep that case statement in 1 place.


bennn:
Merged into https://github.com/racket/scribble/commit/5505a5557ad05cb1dbb44a27ac5122ec5665a82b

Issue #44, opened 2016-06-22
Re-provided identifiers are not tracked across submodules


jsmaniac:
When a library requires and re-provides an identifier without modification, it is normally linked back to the original in documentation for files that require the library `for-label`. This behaviour is however broken when the chain of `require` and re-`provide` is going through a submodule. To test this, install the following package:

```
raco pkg install https://github.com/jsmaniac/example-bug-track-submodule-issue43.git#40c26708c8bec0041a1df444775d6f55161b9e5a
```

The package consists of three files:

**`other.rkt`:**
```racket
#lang racket/base
(require syntax/stx)
(provide (all-from-out syntax/stx))
```

**`main.rkt`:**
```racket
#lang racket/base
(require "other.rkt")
(provide (all-from-out "other.rkt"))
```

**`scribblings/example-bug-track-submodule.scrbl`:**
```racket
#lang scribble/manual
@(require (for-label example-bug-track-submodule
                     racket/base))
@title{example-bug-track-submodule}
@author{georges}
@defmodule[example-bug-track-submodule]

The @racket[stx-map] function is re-provided from @racketmodname[syntax/stx].
```

On the last line of the scribble file, the `@racket[stx-map]` is not linked to the `syntax/stx` library's documentation, and causes the following warning:

```
raco: --- building documentation ---
raco: 1 running: <pkgs>/example-bug-track-submodule/scribblings/example-bug-track-submodule.scrbl
raco: WARNING: undefined tag in <pkgs>/example-bug-track-submodule/scribblings/example-bug-track-submodule.scrbl:
raco:  ((lib "example-bug-track-submodule/main.rkt") stx-map)
```

If the code in `other.rkt` is moved out of the submodule, and `main.rkt` is adjusted accordingly, then the warning disappears and the `@racket[stx-map]` becomes a link to the `syntax/stx` library's documentation. This can be tested using the `without-submod` branch of that package:

```
raco pkg remove example-bug-track-submodule
raco pkg install https://github.com/jsmaniac/example-bug-track-submodule-issue43.git#9747adce0f66bf2a60b58f86b51a0d39c24c0d4b
```

PR #50, opened 2016-08-04
Fixes bug #44 (Re-provided identifiers are not tracked across submodules)

jsmaniac:


bennn:
Looks good

jsmaniac:
I tested this more properly (by installing minimal racket, this patched version of scribble, then the full `main-distribution` package on top of that), and it seems to work fine.

The [example module](https://github.com/jsmaniac/example-bug-track-submodule-issue43.git#40c26708c8bec0041a1df444775d6f55161b9e5a) from bug #44 which re-provides `stx-map` via a submodule works fine with this PR (the identifier `stx-map` is correctly hyperlinked to its original definition, whereas it was underlined in red before).

jsmaniac:
@bennn Just a quick ping, could this be merged before the branch for 6.7 is created (I saw on IRC it should happen tomorrow)?

I've been dog-feeding this patch for a while now, and it works fine. Without it, the documentation for some of my modules is covered with identifiers typeset with a red underline, e.g.:

http://docs.racket-lang.org/phc-toolkit/Syntax_object_manipulation_utilities.html#%28part._.Transformers_utilities%29

On my machine, with this patch, the document displays fine without red hyperlinks, so I was hoping this could be fixed with the next release.

Thanks!

stamourv:
Merged. Sorry for the delay!

PR #52, opened 2016-08-08
use xcolor package (instead of color)

bennn:
`0.` There's already a fix ([`latex-defaults+replacements`](http://docs.racket-lang.org/scribble/core.html#%28def._%28%28lib._scribble%2Flatex-properties..rkt%29._latex-defaults%2Breplacements%29%29)) for my problem, so we can close this if there's any question. But,

afaikt, the [xcolor](https://www.ctan.org/pkg/xcolor?lang=en) package is an improvement over the [color](https://www.ctan.org/pkg/color?lang=en) package:
- provides "basic facilities of `color`" (according to the docs)
- other packages e.g. `tikz` load `xcolor` anyway; when this happens you might get pages of "incompatible color definition on line N" like I've been getting

### DIFF
--- a/scribble-lib/scribble/scribble-load.tex
+++ b/scribble-lib/scribble/scribble-load.tex
@@ -11,7 +11,7 @@
 \newcommand{\packageTextcomp}{\usepackage{textcomp}}
 \newcommand{\packageFramed}{\usepackage{framed}}
 \newcommand{\packageHyphenat}{\usepackage[htt]{hyphenat}}
-\newcommand{\packageColor}{\usepackage[usenames,dvipsnames]{color}}
+\newcommand{\packageColor}{\usepackage[usenames,dvipsnames]{xcolor}}
 \newcommand{\doHypersetup}{\hypersetup{bookmarks=true,bookmarksopen=true,bookmarksnumbered=true}}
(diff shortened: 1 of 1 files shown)


mflatt:
I worry that using "xcolor" instead of "color" will create Latex package conflicts – different ones than you get with "color", which could break existing documents. Is there a worthwhile migration path that doesn't break documents?

rfindler:
As a note, pacm uses xcolor (I think) already too.

bennn:
@mflatt can you send me any documents you think might break? I can try building them.

I think using `latex-defaults+replacements` is an alternative. (I don't think I successfully used it, but I could try again & use xcolor as an example in the docs.)

bennn:
(I think Robby is right.) From http://www.sigplan.org/Resources/Author/ :

> Papers submitted to SIGPLAN events occurring after August 2017 should use the acmart format

And [`acmart.cls`](http://www.sigplan.org/sites/default/files/acmart/current/acmart.cls) uses `xcolor`.

dvanhorn:
FWIW, I have an acmart implementation that I'll issue a PR for soon.  I ran into conflicts with the default scribble style file because of the use of color.

bennn:
Closing. New scribble languages can always do what `scribble/acmart` does and override `\packageColor`.

PR #56, opened 2016-09-16
add HTML 5 sections elements

bennn:
The ones documented here:
http://w3c.github.io/html/sections.html#sections


soegaard:
Hi Benn,

Thanks for the pull request. Adding missing constructs needs
to be done. The question is what the best way of doing this is.

Eli made some good points for a similar pull request:

     https://github.com/racket/scribble/pull/39

Would you be interested in making a scribble/html5 or a scribble/html/5 
with all (or almost all) html5 constructs?

Having these constructs in a separate file avoids various potential problems.

/Jens Axel

bennn:
Thanks for this note -- I should've looked at the other open PR's.

Sure, I'll setup a setup a pull request with `scribble/html5`.

PR #62, opened 2016-11-01
Closes #61

### LINKED ISSUE #61: HTML not responsive enough
The generated HTML is not responsive, IR won't look good on mobile devices.

SpooikypokDev:
Closes #61. Adds better mobile phone support for the generated HTML. Menus are atop the page. Maybe find a better way to do that.

### DIFF
--- a/scribble-lib/scribble/scribble.css
+++ b/scribble-lib/scribble/scribble.css
@@ -482,3 +482,19 @@ ol ol ol ol { list-style-type: upper-alpha; }
     margin-left: 0;
   }
 }
+
+@media (max-width: 480px) {
+  .tocset,
+  .navsettop,
+  .navsetbottom {
+      display: block;
+      width: 100%;
+    }
+  .maincolumn {
+    width: 100%;
+    margin-right: 0px;
+    margin-left: 0px;
+    padding-top: 10px;
+    text-align: justify;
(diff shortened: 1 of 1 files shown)


samth:
@mbutterick is probably the right person to review this.

mbutterick:
I would reject this PR. I see the issue referenced in #61. But this patch does not fix it.

More broadly, I don't think it’s worthwhile to try to make the docs work on mobile phones. The code samples can't be reduced in any reasonable way. For instance, [Rust](https://doc.rust-lang.org) wraps the code samples (which looks terrible). [Go](https://golang.org/doc/) and [Swift](https://developer.apple.com/reference/swift/string) put them in horizontally scrolling boxes (which looks terrible too). 

Racket does nothing, which may also look terrible, but at least keeps us competitive.

But I'll make you a deal: post an example of programming-language documentation that does a better job, and I'll reconsider the matter.

SpooikypokDev:
Godot docs: http://docs.godotengine.org/en/stable/index.html

mbutterick:
These docs don't have the same problem because they don't rely on code samples to the same degree. But the few that exist are in horizontally scrolling boxes. If I have more time to think about this down the line, I will. But for now, case closed.

PR #64, opened 2016-11-04
updated Source Code Pro fonts

mbutterick:
Updated fonts that address alignment problems noticed by @soegaard, for instance in [`2d`](http://docs.racket-lang.org/2d/):

screen shot 2016-11-03 at nov 03 5 59 15 pm

Updated fonts should display like so:

screen shot 2016-11-03 at nov 03 6 01 17 pm


mbutterick:
PS There’s nothing to be done about the italic case, because we omit the true italic styles of the font family to reduce download size. OTOH, maybe that's exactly what an italic border should look like.

screen shot 2016-11-03 at nov 03 6 03 59 pm

rfindler:
Thanks!

PR #75, opened 2017-01-22
Use XeLatex instead of PdfLatex

sorpaas:
It just works, and with the benefits to support unicode characters so I can input Chinese.

I also need the additional styles:

```tex
\usepackage{xeCJK}
\setCJKmainfont{Source Han Sans CN}
```


bennn:
Thanks for the contribution.

IMO, it would be better to add a new flag (`--xelatex` ?) than change the behavior of `--pdf`.  Merging this PR may cause problems with existing documents ([tex.stackexchange](http://tex.stackexchange.com/questions/2984/frequently-loaded-packages-differences-between-pdflatex-and-xelatex)).


(You should be able to use `xelatex` and the additional styles without making changes to the internals of Scribble. First make a style file (`texstyle.tex`) with the `\usepackage` and `\setCJmainfont` commands , then run `scribble ++style texstyle.tex --latex ....`, and finally run `xelatex` on the `.tex` file that Scribble generates.)

sorpaas:
@bennn I updated this with the new flag, `--xelatex`. Have a look.

scribble-lib/scribble/private/run-xelatex.rkt
```diff
@@ -0,0 +1,95 @@
+#lang scheme/base
```
bennn: Can you re-use the code from `run-pdflatex.rkt` instead of copying it?

scribble-lib/scribble/run.rkt
```diff
@@ -67,6 +68,9 @@
    [("--pdf") "generate PDF-format output (via PDFLaTeX)"
     (current-html #f)
     (current-render-mixin pdf:render-mixin)]
+   [("--xelatex") "generate PDF-format output (via XeLatex)"
```
bennn: `XeLaTeX`

scribble-lib/scribble/xelatex-render.rkt
```diff
@@ -0,0 +1,17 @@
+#lang scheme/base
```
bennn: Same here, can you try re-using code from `pdf-render.rkt`? There's like a 6-character diff between the `*-render.rkt` files.

(This is just my opinion, trying to give constructive suggestions until someone with authority to say "yes/no to merge" reviews.)

sorpaas:
Revised and now it reuses as many code as possible.

scribble-lib/scribble/private/run-pdflatex.rkt
```diff
@@ -2,16 +2,21 @@
...
+    (run file notify 'dvipdf)))
+(define (run-xelatex file [notify void])
+  (parameterize ([function-name 'run-xelatex])
+    (run file notify 'xelatex)))
 
 (define max-runs 5)
-(define (run file notify via-dvipdf?)
-  (define latex-cmd-name (if via-dvipdf? "latex" "pdflatex"))
+(define (run file notify type)
+  (define latex-cmd-name (cond [(equal? type 'pdflatex) "pdflatex"]
```
bennn: minor point: since these are symbols you can use `eq?` or `(case type [(pdflatex) "pdflatex"] ....)`

scribble-lib/scribble/private/run-pdflatex.rkt
```diff
@@ -2,16 +2,21 @@
...
+                               [(equal? type 'dvipdf) "latex"]
+                               [(equal? type 'xelatex) "xelatex"]))
```
bennn: add an `else` branch that raises an exception?

scribble-lib/scribble/run.rkt
```diff
@@ -67,6 +67,9 @@
    [("--pdf") "generate PDF-format output (via PDFLaTeX)"
     (current-html #f)
     (current-render-mixin pdf:render-mixin)]
+   [("--xelatex") "generate PDF-format output (via XeLaTex)"
```
bennn: oh, `XeLaTeX` needs to end with a capital `X`

bennn:
@stamourv  ping. Can you review this?

mflatt:
Squashed and merged, and then added docs - thanks!

PR #77, opened 2017-01-24
Handle Chinese Hanzi, Hiragana and Katakana in tag

sorpaas:
Otherwise, any unicode will be replaced by "___", which results in duplicate tags.

### DIFF
--- a/scribble-lib/scribble/base.rkt
+++ b/scribble-lib/scribble/base.rkt
@@ -44,7 +44,7 @@
 
 (define (gen-tag content)
   (datum-intern-literal
-   (regexp-replace* "[^-a-zA-Z0-9_=]" (content->string content) "_")))
+   (regexp-replace* #px"[^-a-zA-Z0-9_=\u4e00-\u9fa5\u3040-\u309F\u30A0-\u30FF]" (content->string content) "_")))
 
 (define (prefix->string p)
   (and p (if (string? p) 
--- a/scribble-lib/scribble/base.rkt
+++ b/scribble-lib/scribble/base.rkt
@@ -44,7 +44,7 @@
 
 (define (gen-tag content)
   (datum-intern-literal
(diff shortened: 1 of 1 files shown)


scribble-lib/scribble/base.rkt
```diff
@@ -44,7 +44,7 @@
 
 (define (gen-tag content)
   (datum-intern-literal
-   (regexp-replace* "[^-a-zA-Z0-9_=]" (content->string content) "_")))
+   (regexp-replace* #px"[^-a-zA-Z0-9_=\u4e00-\u9fa5\u3040-\u309F\u30A0-\u30FF]" (content->string content) "_")))
```
bennn: Just curious, why does the Hanzi range stop at `\u9fa5` instead of `\u9fcc` or `\u9fff`? 

(I don't know Chinese, I just looked up the ranges in a unicode table & on [stack overflow](http://stackoverflow.com/questions/1366068/whats-the-complete-range-for-chinese-characters-in-unicode))

scribble-lib/scribble/base.rkt
```diff
@@ -44,7 +44,7 @@
 
 (define (gen-tag content)
   (datum-intern-literal
-   (regexp-replace* #px"[^-a-zA-Z0-9_=\u4e00-\u9fa5\u3040-\u309F\u30A0-\u30FF]" (content->string content) "_")))
+   (regexp-replace* "[^-a-zA-Z0-9_=\u4e00-\u9fa5\u3040-\u309F\u30A0-\u30FF]" (content->string content) "_")))
```
mflatt: Keeping `#px` is better, because it means that the regexp will be compiled just once, instead of every time that `regexp-replace` is called.

mflatt:
Thanks! I've merged this in squashed form with some extra commentary.

PR #78, opened 2017-01-25
Support hidden elements

sorpaas:
Sometimes you just want an element to be completely hidden. This pull request add hidden element render support for text, markdown, latex and html through a new style tag `'hidden`.

### DIFF
--- a/scribble-lib/scribble/html-render.rkt
+++ b/scribble-lib/scribble/html-render.rkt
@@ -1592,6 +1592,7 @@
            [(smaller) '([class "Smaller"])]
            [(larger) '([class "Larger"])]
            [(hspace) '([class "hspace"])]
+           [(hidden) '([style "display: none"])]
            [(newline) '()]
            [else (error 'html-render "unrecognized style symbol: ~e" name)])]
         [(string? name) (if style null `([class ,name]))]
--- a/scribble-lib/scribble/latex-render.rkt
+++ b/scribble-lib/scribble/latex-render.rkt
@@ -452,6 +452,9 @@
                   [(superscript) (wrap e "textsuper" #f)]
                   [(smaller) (wrap e "Smaller" #f)]
                   [(larger) (wrap e "Larger" #f)]
+                  [(hidden)
+                   (check-render)
+                   (void)]
                   [(hspace)
                    (check-render)
                    (let ([s (content->string e)])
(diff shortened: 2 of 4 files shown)


mflatt:
@sorpaas Can you say more about the motivation here? Since Scribble content is generated programmatically, it seems like the program could just not generate the element (or generate it with empty content) as easily as it could generate an element with a `'hidden` style.

sorpaas:
First of all, thanks for the review!

I have a project (an org-mode alternative) that takes heavy use of `traverse-element` and `traverse-block`. For example, I may have a `task` function that returns a `traverse-element` who will then insert the task information into `ti`.

```
(define (task t)
  (traverse-element
    (lambda (get set)
      ;;; Use set to insert the information
)))
```

Then I can create a function, for example, `agenda`, that prints out all the tasks in a later traverse phase.

```
(define (agenda)
  (traverse-element
    (lambda (get set)
      (lambda (get set)
        ;;; Print out all the tasks in the previous traverse phase
))))
```

And then I might want to create something, for example, an `event-that-requires-preparation` to reuse `task`. This, if without `'hidden` would be a little complicated. Basically, I want to reuse the traverse element defined by `task` (together with all its traverse phase information), but without showing the result of the traverse element. In this way, the task created by `event-that-requires-preparation` would still be printed out by `agenda`, but would not show up redundantly around when the function is called. With `'hidden`, I can do this:

```
(define (event-that-requires-preparation e)
  (list
    (element 'hidden
      (task (string-append "Prepare " e)))
    (traverse-element
      ;;; Do something else for this event
)))
```

You can argue that I can take out the returned traverse element's procedure using `traverse-element-traverse` and then "inject" that into a central traverse element. However, when there are more than 2 traverse elements to be defined, the code can be unreadable. Have a `'hidden` would just be convenient and can make the problem simple. Plus, `'hidden` looks like a general property that might find its usage in other problems.

mflatt:
Ok, I can see why you'd want `'hidden'. This will be ok with a more complete patch that includes docs.

I think there are some interactions here with link targets, where information is collected for for a target in a `'hidden` element, but the target turns out not to be present; maybe that's ok, and it just needs to be documented. Also, the HTML output seems different than other backends, in that the content is present but made hidden through CSS; maybe it would be more consistent to node generate the output for HTML.

PR #84, opened 2017-03-02
add 'unnumbered' style to footnote parts

bennn:
The function bound to `make-my-notes` by `define-footnote` (as in:

```
  (define-footnote my-note make-my-notes)
```

) now produces a part with the `unnumbered` style.

This way, footnote sections can go between any two sections on a page
 without upsetting the section numbers.


bennn:
I'm also thinking:

- [ ] the docs need an example
- [ ] it would be easier to document a `make-footnote` function that returns 2 values, and have `define-footnote` as a macro on top of that
- [ ] the `footnote-part` function shouldn't take any arguments  ... right? it just ignores them ([link](https://github.com/racket/scribble/blob/master/scribble-lib/scriblib/footnote.rkt#L44))

bennn:
ok I'm happy with this now

mflatt:
LGTM

PR #85, opened 2017-03-04
Scribble support for new acmart.cls.

dvanhorn:


scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+
+@defproc[(maketitle) block?]{
+
+Issues the @tt{\maketitle} command.  This must appear after the abstract
+any several other top-matter commands.  (See the acmart latex documentation.)}
+
+
+@deftogether[(
+@defproc[(acmJournal [journal pre-content?] ...) content?]
+@defproc[(acmConference [name pre-content?] [date pre-content?] [venue pre-content?]) content?]
```
bennn: no `short-name` optional argument?

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+@defproc[(acmVolume [content pre-content?] ...) content?]
+@defproc[(acmNumber [content pre-content?] ...) content?]
+@defproc[(acmArticle [content pre-content?] ...) content?]
+@defproc[(acmYear [content pre-content?] ...) content?]
+@defproc[(acmMonth [content pre-content?] ...) content?]
+@defproc[(acmArticleSeq [content pre-content?] ...) content?]
```
bennn: For `acmVolume acmNumber acmArticle acmYear acmMonth acmArticleSeq`, I think `integer?` is the right contract & they should just take 1 argument.

(Can't think of why someone would want to pass something that doesn't evaluate to an integer)

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+@defproc[(acmPrice [content pre-content?] ...) content?]
```
bennn: Same, I think the domain should be `flonum?`

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+@defproc[(acmISBN [content pre-content?] ...) content?]
+@defproc[(acmDOI [content pre-content?] ...) content?]
```
bennn: For `acmISBN` and `acmDOI` I think the domain should be `pre-content?`

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+@defproc[(acmBadgeL [content pre-content?] ...) content?]
+@defproc[(acmBadgeR [content pre-content?] ...) content?]
```
bennn: For `acmBadgeL` and `acmBadgeR`, I think the domain should be `string? path-string?`
(or at least, the same as `hyperlink`'s domain)

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+@defproc[(department [content pre-content?] ...) content?]
+@defproc[(streetaddress [content pre-content?] ...) content?]
+@defproc[(city [content pre-content?] ...) content?]
+@defproc[(state [content pre-content?] ...) content?]
+@defproc[(postcode [content pre-content?] ...) content?]
+@defproc[(country [content pre-content?] ...) content?]
+)]{
+
+Declares information that is collected for each author.  These commands should
+ only be used within an @racket[affiliation] command.}
```
bennn: Add an example?

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+See also @url["http://www.acm.org/about/class/how-to-use"].
+
+For @racket[terms], each general term should be in titlecase. Terms
+are usually drawn from a fixed list, and they are usually optional.
+
+For @racket[keywords], capitalize only the first letter of the first
+word, separate phrases by commas, and do not include ``and'' before
+the last one. Keywords should be noun phrases, not adjectives.}
+
+@defproc[(startpage [content pre-content?] ...) content?]{
```
bennn: Should it be `startPage`?
(To match the macro I see in the acmart v1.28 docs)

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+
+@defproc[(ccsdesc [content pre-content?] ...) content?]{
+
+Declares CCS description.}
+
+@defproc[(received [content pre-content?] ...) content?]{
+
+Sets the date the paper was received.}
+
+@defproc[(citestyle [content pre-content?] ...) content?]{
```
bennn: Does this need the `...` ?

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+@defproc[(printonly [content pre-content?] ...) block?]
+@defproc[(screenonly [content pre-content?] ...) block?]
+@defproc[(anonsuppress [content pre-content?] ...) block?]
+)]{
+Marks content to be included only for print or screen
+editions, or excluded from anonymous editions.}
+
+@defproc[(acks [content pre-content?] ...) block?]{
+
+Creates an acknowledgments section.}
```
bennn: unlabeled? or does it do what acmart's `ack` does?

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
+#lang scribble/manual
+@(require "utils.rkt" (for-label (except-in scribble/acmart title author)))
+
+
+@title{ACM Paper Format}
+
+@defmodulelang[scribble/acmart]{The @racketmodname[scribble/acmart]
+language is like @racketmodname[scribble/base], but configured with
+Latex style defaults to use the @filepath{acmart.cls} class
```
dvanhorn: Just like `sigplan`, but I'll change both.

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+@defacmartid[sigconf]
+@defacmartid[siggraph]
+@defacmartid[sigplan]
+@defacmartid[sigchi]
+@defacmartid[sigchi-a]
+@defacmartid[review]
+@defacmartid[screen]
+@defacmartid[natbib]
+@defacmartid[anonymous]
+@defacmartid[authorversion]
```
dvanhorn: OK.

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+The @racket[manuscript], @racket[acmsmall], @racket[acmlarge], 
+@racket[acmtog], @racket[sigconf], @racket[siggraph], @racket[sigplan],
+@racket[sigchi], and @racket[sigchi-a] options are all mutually exclusive,
+but may each be used in combination of @racket[review], @racket[screen], @racket[natbib],
+@racket[anonymous], and @racket[authorversion].
+
+
+@defproc[(abstract [pre-content pre-content?] ...) block?]{
+
+Generates a @tech{nested flow} for a paper abstract.}
```
dvanhorn: It generates exactly `\begin{syntax}`.

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+
+@defform[(include-abstract module-path)]{
+
+Similar to @racket[include-section], but incorporates the document in the
+specified module as an abstract. The document must have no title or
+sub-parts.}
```
dvanhorn: This is just like the existing sigplan library.

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -0,0 +1,157 @@
...
+
+@defproc[(subtitle [pre-content pre-content?] ...) content?]{
+
+Use as the last argument to @racket[title] to specify a subtitle.}
+
+@defproc[(maketitle) block?]{
+
+Issues the @tt{\maketitle} command.  This must appear after the abstract
+any several other top-matter commands.  (See the acmart latex documentation.)}
```
dvanhorn: 1. Yes.
2. Yes.
3. I think that would be a mess.  The better thing would be to automatically insert `maketitle` at the right point, but I couldn't figure out how to do it.

scribble-doc/scribblings/scribble/acmart.scrbl
dvanhorn: What is the idiomatic way of supporting optional arguments in LaTeX commands?  Best I could come up with was manually define two latex wrapper commands.

scribble-doc/scribblings/scribble/acmart.scrbl
dvanhorn: It's easy to write strings in scribble and doesn't require remembering the right type to use.  This interface is more consistent with the latex interface _and_ the existing `sigplan` library.

scribble-doc/scribblings/scribble/acmart.scrbl
dvanhorn: Same response.

scribble-doc/scribblings/scribble/acmart.scrbl
dvanhorn: Tried to be consistent with `sigplan`.

scribble-doc/scribblings/scribble/acmart.scrbl
dvanhorn: I've switched it to `string?` (and will add the optional argument when I know how to).

scribble-doc/scribblings/scribble/acmart.scrbl
dvanhorn: OK.

scribble-doc/scribblings/scribble/acmart.scrbl
dvanhorn: Does what `ack` does.

scribble-doc/scribblings/scribble/acmart.scrbl
bennn: Ah! that explains a lot. (I had never read the `scribble/sigplan` docs.)

scribble-doc/scribblings/scribble/acmart.scrbl
bennn: Good question, I don't know either.

dvanhorn: I added the optional URL argument for `acmBadgeL` and `acmBadgeR` and can do the same for these (but I'll first see if there's a better way to do this).

dvanhorn:
I've taken care of all the FIXMEs (but one) and I believe addressed all your comments.  The last remaining issue is what to do about the theorem styles.  Not sure if they should be included in the library or not.

mflatt:
When I try this, I get `cannot open input file "acmart-load.tex"`. Is that file missing?

mflatt:
Also, doc rendering reports
```
raco setup: WARNING: undefined tag in <pkgs>/scribble-doc/scribblings/scribble/scribble.scrbl:
raco setup:  ((lib "scribble/acmart.rkt") email)
```

dvanhorn:
I fixed both of these issues, @mflatt.  Thanks!

mflatt:
@dvanhorn I still get the same error, but changing `"acmart-load.tex"` to `(scribble-file "acmart/acmart-load.tex")` now fixes that much.

But then I get `! LaTeX Error: \SConferenceInfo undefined.` for `\renewcommand{\SConferenceInfo}`. Do I need to try something more than
```
#lang scribble/acmart
@title{Example}
```
?

dvanhorn:
It looks like you at least need an abstract, but I'm not sure why.  I'll try to fix.

mflatt:
It looks like `title` is also needed before `maketitle` – not surprising, but probably worth mentioning in the docs. Also, if there's no good way around having to use `maketitle`, probably `maketitle` should just be mentioned at the beginning of the `scribble/acmart` docs.

dvanhorn:
@mflatt, your example now works once `maketitle` is added (I wasn't using the style properties pervasively that triggered the tex-addition).  Will document `maketitle` earlier and more prominently.

dvanhorn:
I moved `maketitle` up, included a minimal example, and specified that it is mandatory.

PR #87, opened 2017-03-11
defstruct: no newline between name & fields if no fields

bennn:
If a struct has no fields, don't check if a struct's name + first field's name
are too long to print on one line

Bug report: https://groups.google.com/forum/#!topic/racket-users/6fKGky4tud8


stamourv:
LGTM

PR #88, opened 2017-03-15
add font size options to scribble/acmart

bennn:
Just like the 10pt option in `scribble/sigplan`,
 putting @9pt @10pt @11pt or @12pt on the #lang line of a `scribble/acmart`
 document passes to corresponding string to `\documentclass[....]`

And just like TeX, it's ok to give multiple font sizes. All but the last are ignored.

(DLS for example want `sigplan` format and 10pt font. I added all the options because, why not.)


dvanhorn:
These aren't documented options for acmart, but the package does seem to accept them.  Since DLS is asking for their use it seems to make sense to add them.

PR #92, opened 2017-03-21
Last,First format for names in bibtex

jsmaniac:
Handles comma-separated Last,First and Last,Jr,First format for names in bibtex, as well as `von`, `de la` and `van der` separators.

Note that this patch does not add support for disambiguating braces as bibtex normally supports, e.g. `{first fst} {last lst}`. That would require a more involved parser.

### DIFF
--- a/scribble-lib/scriblib/bibtex.rkt
+++ b/scribble-lib/scriblib/bibtex.rkt
@@ -225,8 +225,9 @@
               (author-name (trim one) (trim two) #:suffix (trim suffix))]
              [(pregexp #px"^(.*),(.*)$" (list _ two one))
               (author-name (string-trim one) (string-trim two))]
-             [(pregexp #px"^(.*) (von|de la|van der) (.*)$" (list _ one von-like two))
-              (author-name (string-trim one) (string-append von-like " " (string-trim two)))]
+             [(pregexp #px"^(.*?)\\s+(\\p{Ll}[^\\s]*(\\s+\\p{Ll}[^\\s]*)*)\\s+(.*)$" (list _ one von-like _ two))
+              (author-name (string-trim one)
+                           (string-append (string-trim von-like) " " (string-trim two)))]
              [space-separated
               (match (regexp-split #px"\\s+" space-separated)
                 [(list one) (org-author-name one)]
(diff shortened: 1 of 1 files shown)


jsmaniac:
Note: Excluding the support for braces, I followed the syntax for names as described here: http://tex.stackexchange.com/questions/557/how-should-i-type-author-names-in-a-bib-file

jsmaniac:
Hm, it seems I have introduced somehow a regression, `Chang, Knauth and Greenman@~cite{chang2017type-systems-as-macros}` gives me `bibtex: Unknown citation "chang2017type-"`. Let me investigate this, I'll post back soon.

jsmaniac:
Fixed, the bug was simply that `\\s` requires `#px`, and the regexps were using `#rx`, so `\\s` was interpreted as a plain `s`, and `~cite` was splitting on occurrences of `s`, not on spaces :) .

The PR should be good now.

bennn:
re: `von` etc, I do think you should change to accept any word that starts with a lowercase letter. I guess you found some docs, and I found this while reading your stackexchange link

> BibTeX knows where one part ends and the other begins because the tokens in the von part begin with lower-case letters.

http://nwalsh.com/tex/texhelp/bibtx-23.html

After that, I think we should merge 👍

jsmaniac:
@bennn Done!

bennn:
Word! I sent Matthew a ping on slack to look --- because I'm nervous about seizing authoring and clicking the "merge" button. But if he's busy then I'll summon the courage tomorrow morning.

One nagging question in the meantime: should all the `.*` be `[^\\s]*`?

mfelleisen:
@bennn push Ben push!

jeapostrophe:
Amen

jsmaniac:
@bennn Thanks for merge :) . The `.*` actually do and should catch spaces, e.g. in `Last L, First Name`. The ambiguity arises when there are more than three commas, `a,b,c,d`, the regexp will currently treat it as last=`a,b` junior=`c` first=`d`, but I'm not sure if that's how BibTeX handles it (or if BibTeX sees it as last=`a` junior=`b` first=`c,d`, or some other combination). Putting more than three commas is asking for trouble anyway, and should not happen.

PR #94, opened 2017-03-25
Add a pretitle style property for nested flows.

LeifAndersen:
This allows us to raise nested flows above the title. So that we do
things like raise the abstract above the title:

```
\begin{abstract}
Abstract text
\end{abstract}
\titleCommand{...}
```

This style is required by the acmart style guide.

(Documentation will be in the next commit.)

### DIFF
--- a/scribble-doc/scribblings/scribble/core.scrbl
+++ b/scribble-doc/scribblings/scribble/core.scrbl
@@ -729,6 +729,8 @@ The following @tech{style properties} are currently recognized:
  @item{@racket[alt-tag] structure --- Generates the indicated HTML tag
        instead of @tt{<blockquote>}.}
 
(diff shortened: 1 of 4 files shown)


mflatt:
Looks generally ok to me.

Probably the docs should say more about the interaction of the paragraph `'pretitle` style and the nested-flow `'pretitle` style property, where (currently) all of the paragraphs precede all of the nested flows. Or maybe it would be better to run a single extraction pass to keep paragraphs and nested flows in the same relative order.

LeifAndersen:
Thanks for the feedback. That might make sense, although then we'd have to use a different function to extract flows and paragraphs at the same time. Although that seems easy enough. I'll give that a go.

Thanks again.

Issue #99, opened 2017-04-03
Scribble will split @racket[...] identifiers down the middle.


LeifAndersen:
If I have the following scribble program:

```
#lang scribble/sigplan

@require[scribble/manual]

Paragraph
@racket[(some-racket-function name with arguments)]. This
paragraph also has a bunch of text so that we can get a good
sense of what the fill box is.
```

When I render it to a pdf, the `arguments` identifier gets split across to a new line, making:

![screen shot 2017-04-03 at 5 03 05 pm](https://cloud.githubusercontent.com/assets/285926/24631734/70573b76-188f-11e7-9855-1403e55dd5bf.png)

I get that a chunk of `@racket[...]` code is in the text and might wrap, but splitting it in the middle of an identifier like this seems bad.

jeapostrophe:
In particular, it makes it seem like the argument is "ar-guments" which could be bad if a function name were rendered wrong, etc.

Leif, can you check where the - is inserted? Is it by the LaTeX backend? If so, we could put code blocks in `\nohyphens{}`

LeifAndersen:
I strongly suspect its inserted by latex, and I'll see if is fixable. I agree that something like `\nohyphens{}` is the right way to go about fixing this.

LeifAndersen:
Also, given that the file expands to the following tex code, I think its safe to say that its latex inserting it:

```
Paragraph
\RktPn{(}\RktSym{some{-}racket{-}function}\Scribtexttt{ }\RktSym{name}\Scribtexttt{ }\RktSym{with}\Scribtexttt{ }\RktSym{arguments}\RktPn{)}. This
paragraph also has a bunch of text so that we can get a good
sense of what the fill box is.
```

LeifAndersen:
Okay, adding `\nohyphens{}` causes the output to be:
![screen shot 2017-04-03 at 6 18 29 pm](https://cloud.githubusercontent.com/assets/285926/24634102/009434dc-189a-11e7-96f3-6672d459df02.png)

At least no hyphens are created, although I wish it would take the identifier to the next line.

LeifAndersen:
Moving conversation to PR #100

PR #100, opened 2017-04-03
Prevent hyphens from appearing in @racket[...] identifiers when they …

LeifAndersen:
…overflow.

Fixes issue #99

### DIFF
--- a/scribble-doc/scribblings/scribble/config.scrbl
+++ b/scribble-doc/scribblings/scribble/config.scrbl
@@ -640,6 +640,20 @@ style:
 
  ]
 
+Additionally, the @filepath{racket.tex} Latex configuration
+includes several macros that you can redefine to adjust the
+output style of Racket code:
+
+@itemlist[
+ @item{@ltxd[2]{SColorize} --- Sets the color scheme of
+  Racket code. Can be redefined to create black and white
+  code.}
+ @item{@ltxd[1]{SHyphen} --- Enables or Disables the ability
+  for identifiers and keywords in Racket code from being
+  hyphenated. Defaults to enabled (for compatibility).
+  Redefine to disable, for example:
+  @tt["\\renewcommand{\\SHyphen}[1]{\\mbox{#1}}"]}]
+
 @; ------------------------------------------------------------
 
 @section[#:tag "latex-prefix"]{Latex Prefix Support}
--- a/scribble-doc/scribblings/scribble/config.scrbl
+++ b/scribble-doc/scribblings/scribble/config.scrbl
@@ -647,12 +647,17 @@ output style of Racket code:
 @itemlist[
  @item{@ltxd[2]{SColorize} --- Sets the color scheme of
   Racket code. Can be redefined to create black and white
-  code.}
+  code. The first argument is the requested color, and the
+  second argument is the text for that color.}
  @item{@ltxd[1]{SHyphen} --- Enables or Disables the ability
   for identifiers and keywords in Racket code from being
   hyphenated. Defaults to enabled (for compatibility).
-  Redefine to disable, for example:
-  @tt["\\renewcommand{\\SHyphen}[1]{\\mbox{#1}}"]}]
+  Redefine to disable or change the hyphenation behavior. For
+  example, to cause the text to overfill rather than hyphen,
+  it can be redefined to:
+  @tt["\\renewcommand{\\SHyphen}[1]{\\mbox{#1}}"]. The first
+  argument is an identifier or keyword inside of a code
+  block.}]
 
 @; ------------------------------------------------------------
 
--- a/scribble-lib/scribble/racket.tex
+++ b/scribble-lib/scribble/racket.tex
@@ -2,7 +2,10 @@
 % Redefine \SColorize to produce B&W Scheme text
 \newcommand{\SColorize}[2]{\color{#1}{#2}}
 
-\newcommand{\inColor}[2]{{\nohyphens{\Scribtexttt{\SColorize{#1}{#2}}}}}
+% Redefine SHyphen to allow identifiers to be hyphenated
+\newcommand{\SHyphen}[1]{\nohyphens{#1}}
+
+\newcommand{\inColor}[2]{{\SHyphen{\Scribtexttt{\SColorize{#1}{#2}}}}}
 \definecolor{PaleBlue}{rgb}{0.90,0.90,1.0}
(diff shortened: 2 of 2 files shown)


rfindler:
This seems like a big change, and I'd worry that it might break other things.

I don't have any concrete issues, but the change seems a bit scary to me. Not just because I think that it is okay to have hyphens (and, indeed, preferable in some cases) but also because this macro is a pretty low-level one.

I think that maybe it might be good to add an optional keyword argument to @racket[] and perhaps have a default value for it that could be set on a per-document basis?

LeifAndersen:
> This seems like a big change, and I'd worry that it might break other things.

Ya, I get that fear. I have no idea if this would break a lot of things as there really isn't a set of good scribble tests (Yet, that PR will come eventually. :) ), but I can say it does work for my documents.

> Not just because I think that it is okay to have hyphens (and, indeed, preferable in some cases) but also because this macro is a pretty low-level one.

Huh, interesting. Can you give me a case where you would want a hyphen in the middle of an identifier or keyword? (That wasn't already there anyway. I'd be okay with doing it for identifiers that already have them.)

> I think that maybe it might be good to add an optional keyword argument to @racket[] and perhaps have a default value for it that could be set on a per-document basis?

I'm not opposed to having an optional argument. Although I would be opposed to needing to set it for every @racket[] block. I'd be down for a document wide setting, but how would you do that? (A parameter is a good start, but it would need to be set in every module.)

rfindler:
I think the reason that one might want a hyphen is the same reason that one might want a hyphen in ordinary words. I do agree with you that usually it is not what you want. But we also don't want to go into the margin and it isn't obvious which badness is worse.

As for document-wide setting, I guess one could use state. But maybe it should be delimited, so a parameter might be better.

LeifAndersen:
> But we also don't want to go into the margin and it isn't obvious which badness is worse.

That is true. IMO the best thing to do is to underfil the line, but even that is up for debate.

> As for document-wide setting, I guess one could use state. But maybe it should be delimited, so a parameter might be better.

So I added the SHyphen macro, which is like the SColorize one defined in the same file. That way there is a hook that writers can use to set how they want identifiers to be handled.

mflatt:
I think the default shouldn't change. I can appreciate wanting to disable hyphenation everywhere, but it's a trade-off between a kind of precision (not including a "-" that might be mistaken as part of the identifier) and other concerns (like intraline spacing), and we shouldn't change existing documents.

LeifAndersen:
That's fair. We could add `SHyphen` as a no-op, so that style files can define it as mbox or hyphen if they want, how does that sound?

rfindler:
I agree with Matthew.

Thank you for making it easy to customize this behavior, tho! Would it make sense to use this as an example to make the docs better so people have more customization examples to work from? I think some code that starts by saying "here's what yo uneed to add to your @title{} to get rid of these hyphens" could be generally very useful to people as they could work from that other kinds of customizations at that level. (We need more help for people at that level in Scribble.)

LeifAndersen:
Ya sure. I also agree that it can be an example to the docs, or it could also be a tutorial in blog post form (or both.)

mflatt:
Adding `\SHyphen` as a documented no-op sounds good.

LeifAndersen:
Okay cool. I'll make that change, add documentation, create a tutorial for the blog.

LeifAndersen:
Okay, I've added documentation. I'll merge this and write up a short tutorial as a blog post. Unless anyone has any last minute thoughts or issues? (cc @rfindler and @mflatt )

scribble-doc/scribblings/scribble/config.scrbl
```diff
@@ -640,6 +640,20 @@ style:
 
  ]
 
+Additionally, the @filepath{racket.tex} Latex configuration
```
bennn: LaTeX

scribble-doc/scribblings/scribble/config.scrbl
```diff
@@ -640,6 +640,20 @@ style:
...
+includes several macros that you can redefine to adjust the
+output style of Racket code:
+
+@itemlist[
```
bennn: Both these should say what they're called with (rendered arguments of `@racket` ?) so it's clearer how to override them. Reading about `SHyphen`, it seemed like a boolean parameter until I got to the example.

scribble-doc/scribblings/scribble/config.scrbl
LeifAndersen: I have no opinion one way or the other. But the rest of the docs say Latex. So if we do make it LaTeX, we should change that consistently.

bennn: Today can be the first day of the rest of this document's life.

LeifAndersen: Let me rephrase...this should be a second PR. ;)

scribble-doc/scribblings/scribble/config.scrbl
LeifAndersen: Thanks, fixed.

scribble-doc/scribblings/scribble/config.scrbl
```diff
@@ -640,6 +640,25 @@ style:
...
+  second argument is the text for that color.}
+ @item{@ltxd[1]{SHyphen} --- Enables or Disables the ability
+  for identifiers and keywords in Racket code from being
+  hyphenated. Defaults to enabled (for compatibility).
+  Redefine to disable or change the hyphenation behavior. For
+  example, to cause the text to overfill rather than hyphen,
+  it can be redefined to:
+  @tt["\\renewcommand{\\SHyphen}[1]{\\mbox{#1}}"]. The first
+  argument is an identifier or keyword inside of a code
+  block.}]
```
bennn: I still don't think this documentation is clear. Here's an idea:
- remove `\SHyphen`, make a `\newif\ifSHyphen` and change `\incolor` to use the if
- export & document `\SHyphentrue` and `\SHyphenfalse`
(This only makes sense if there's just one way to enable / disable hyphens)

LeifAndersen: Ah, okay. No, I'd rather leave this as is because while it makes sense to enable/disable hyphens, there's lots of different strategies that may or may not be okay for your document. You might be okay with, say, text box overflows, or you might prefer underflows, or you might want an abbreviation (although I admit I'm stretching on that one.)

LeifAndersen:
Okay, since it seems to be okay, I'm going to merge this in.

PR #107, opened 2017-05-15
doc: point 'tabular #:style' to 'table' docs

bennn:
@mflatt thanks for the help on slack, `#:style 'block` does produce a LaTeX `tabular`

I think this PR fixes a typo.

If it's not a typo, I'll change it to a "see also". Because the docs for `para` didn't help me understand `tabular`'s `#:style`
(If I had the list of styles from `table`'s docs, I could have at least tried those to see what happened.)


mflatt:
It's not a typo. The intent was to connect to the way `para` parses the `#:style` argument (i.e., how it converts to a property style).

It would make sense to point to the `table` entry's list of style and style properties, though.

bennn:
Ok, I put the old sentence back & added a new sentence "See table for .... style ...."

Issue #106, opened 2017-05-14
updating acmart.cls


bennn:
cc @LeifAndersen 

Problem:
1. the true `acmart.cls` sometimes changes (<https://github.com/borisveytsman/acmart>)
2. the `scribble/acmart` folder keeps a copy of `acmart.cls` in version control

Proposed solution: do what `scribble/jfp` does, and
- remove `acmart.cls` from version control
- add code to check whether `scribble/acmart/acmart.cls` exists, and if not re-download it

Is there any reason to not do this?

- - -

(It would be good to have a way of detecting changes in `borisveytsman/acmart` and automatically updating, but this at least this way you can delete your `acmart.cls` to get a new one.)

LeifAndersen:
Hmm...actually, I wonder if we can even have `acmart.cls` in version control due to copyright concerns.

It really should be downloaded by the library, like other scribble packages.

dvanhorn:
I think downloading is the right thing to do, but just to add another wrinkle: `acmart.cls` is itself generated.  If you want to use the source code from Boris, you'll need to run `latex acmart.ins` to generate `acmart.cls`.  The real source file is `acmart.dtx`.

bennn:
Working on this now, I think I will use this URL:
http://www.sigplan.org/sites/default/files/acmart/current/acmart.cls

And make a note in the docs about the GitHub repo and how to build & use that `acmart.cls`.

Sounds ok?

jsmaniac:
@bennn This will prevent building such documents on the pkg-build server (builds are run without network access). AFAIK there is no "list of files to be downloaded" or an equivalent mechanism in `info.rkt` (for now).

I guess packages which include an `acmart` document will be pretty rare, though, but embedding research articles with implementation packages and vice versa should be something to encourage, not prevent.

samth:
In addition to what @jsmaniac says, I believe that we can distribute acmart.cls just fine (it's available under the Latex Public License, which says that redistribution is ok).

bennn:
Then #108 is unnecessary, and we should just keep `acmart.cls` up-to-date manually?

jsmaniac:
@bennn Maybe we could keep a local copy, and try to update it (if the update fails, proceed with the already-downloaded copy). Would it be difficult/problematic to do so?

bennn:
But, when is a good time to update? Every time `scribble/acmart` renders a document?

[...]

PR #108, opened 2017-05-15
remove acmart.cls

bennn:
Add code to download `acmart.cls` if a local copy does not exist.

#106


scribble-lib/scribble/acmart/.gitignore
```diff
@@ -0,0 +1 @@
+acmart.cls
```
bennn: Q. is it ok to have this local `.gitignore`? Or should I move this line to the top-level `.gitignore`?

scribble-lib/scribble/acmart/lang.rkt
```diff
@@ -147,3 +156,18 @@
   (struct-copy part doc [to-collect
                          (cons (terms) ; FIXME
                                (part-to-collect doc))]))
+
+(define (acmart.cls)
```
bennn: I'm thinking about moving this to a helper function with type `string? (-> path-string? void?) -> void?` and sharing between `scribble/jfp`.

bennn:
Rendered docs:

screen shot 2017-05-15 at 14 47 31

bennn:
Closed as "not necessary"

PR #109, opened 2017-05-15
[WIP] Extended support for highlighting

jsmaniac:
This patch:

* Adds support for custom CSS styles for `(code:hilite code style-name)`
* Fixes rendering of `@racket[(a . (b c))]` so that the dot is indeed shown if it were present in the source
* I tentatively added a case when determining the first element `[(code:line . rest) #'rest]`. I'm not sure it's correct, though.
* Adds support for code:hilite in a dotted tail position: `(foo bar code:hilite baz)` renders as `(foo bar . baz)`, with the `baz` highlighted. `(foo bar code:hilite baz style-name)` should also work, and take into account the style name.
* Adds support for dotted tail elements to `code:line`, so that `(code:line a b . c)` renders as `a b . c`, just as `(code:line a b c)` renders as `a b c`.
* Fixes the error message given when a `code:hilite` is misused (previously: `"bad code:redex: ~.s"`, now: `"bad code:hilite: ~.s"`).
* Disabled the spurious `(set! src-col (add1 src-col))` for `code:hilite` when the `syntax-span` of the whole form is `0`. This allows tools to wrap elements with `code:hilite` forms, without having an extra space. The tool needs to give the whole `(code:hilite e)` form the same srcloc as `e`, except for the `syntax-span` which is overridden with `0`.

These features / changes allow tools to highlight parts of the code, for example `sexp-diff` could be used to generate pretty diffs (such a tool would need to take care of srclocs, though), or to show before/after steps in literate programming. I wrote a small utility which does this:

Screenshot showing removed parts in red, added parts in green, existing parts dimmed

The bulk of the diff comes from moving `lloop` out of its former `cond` case, so that it can be used both by the `[(or (pair? (syntax-e c)) …` case and by the `code:line` case.

Issues to solve before merging:
- [ ] I'm not sure if the `[(code:line . rest) #'rest]` is correct (first line of the diff). I think it won't cause any harm, but I'd like one of the scribble devs to double-check it.
- [ ] Documentation
- [ ] It would be nice to be able to specify a highlight style, instead of specifying a style name. Currently, it is necessary to manually inject the `css-additions` outside of the code. I'm not sure how to embed a style without resorting to 3D syntax, or performing a dynamic-require of some sort. Ideas welcome!

### DIFF
--- a/scribble-lib/scribble/racket.rkt
+++ b/scribble-lib/scribble/racket.rkt
@@ -326,6 +326,7 @@
            [first (if escapes?
                       (syntax-case c (code:line)
                         [(code:line e . rest) #'e]
+                      [(code:line . rest) #'rest]
                         [else c])
                       c)]
            [init-col (or (syntax-column first) 0)]
@@ -380,7 +381,7 @@
               (set! content (cons (elem-wrap
                                    ((if highlight?
                                         (lambda (c)
-                                          (make-element highlighted-color c))
+                                          (make-element highlight? c))
                                         values)
                                     (if (and color? cls)
                                         (make-element/cache cls v)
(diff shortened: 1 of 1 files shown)


scribble-lib/scribble/racket.rkt
```diff
@@ -326,6 +326,7 @@
            [first (if escapes?
                       (syntax-case c (code:line)
                         [(code:line e . rest) #'e]
+                      [(code:line . rest) #'rest]
```
bennn: I _think_ this case only matches `(code:line)`, so it shouldn't hurt but also shouldn't change how anything looks.

jsmaniac: With the patch which I added, you can now write `(code:line . foo)` which renders as `. foo` (but previously it was not possible).

Also, technically, we could have a first element which is `(code:line (code:line something))`, should the code above be called recursively as long as `code:line` is the first element? (I can amend the patch to include that, if you think it's the right thing to do).

jsmaniac:
Note to self: The problem when making sure that `@racket[(a . (b c))]` gets rendered as `(a . (b c))` instead of `(a b c)`, i.e. the second point of the PR, is that scribble's `@defform` produces nested syntax pairs, instead of producing a syntax list.

I would therefore either have to "fix" `defform` (and test the other forms, `@defproc` works fine but others may exhibit similar problems), or remove this part of the patch, or make it an option, e.g. accessible via `@racket/dots[(a . (b c))]`.

PR #110, opened 2017-05-16
Handle "last, first" name format in scriblib/bibtex parsing

dvanhorn:
Addresses problem report 15431: if a bibtex entry uses "last, first"
name format, then author-year citations generate references using
first names.

This patch is from Nicholas Labich (@labichn).

### DIFF
--- a/scribble-lib/scriblib/bibtex.rkt
+++ b/scribble-lib/scriblib/bibtex.rkt
@@ -218,9 +218,16 @@
       (apply authors
          (for/list ([a (in-list (regexp-split #rx" +and *" as))])
            (match (regexp-split #rx" +" a)
-             [(list one) (org-author-name one)]
-             [(list one two) (author-name one two)]
-             [(list-rest first rest) 
+             [(list one)
+              (org-author-name one)]
+             [(list (pregexp #px"(.*)," (list _ lst)) fst)
+              (author-name fst lst)]
+             [(list one two)
+              (author-name one two)]
+             [(list-rest (pregexp #px"(.*[^,])" (list _ lstpre)) ... (pregexp #px"(.*)," (list _ lst)) fsts)
+              (author-name (apply string-append (add-between fsts " "))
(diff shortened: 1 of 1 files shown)


LeifAndersen:
Provided no one has a comma as the last character of their first name...LGTM.

(Or if they do have a comma as the last character of their first name, bibtex files are supposed to escape them.)

bennn:
This seems simpler than #92, and passes all the tests in that PR. @jsmaniac do you see an important difference between this and #92 ?

minor point: I'm confused why `parse-author` uses " +" instead of "\\s+"

jsmaniac:
@bennn I don't think this patch will handle `Last,First` (without a space after the comma, I added test cases for this now), but BibTeX and my patch do handle it.

It also seems to lack support fot `von`, `de la`and `van der` (for these, BibTeX seems to actually accept any sequence of words starting with a lowercase letter, e.g. `First fOn bER Last`, but the docs I had found were more restrictive. Should I update my patch to allow any lowercase-starting word instead of the hardcoded `von`, `de la` and `van der`?) .

It also seems to lack support for `Last, Jr, First`.

I'm surprised it passed the (former) tests in #92, do `authors` and `author-name` have a `prop:custom-write` which prevents the smaller distinctions between `(author-name "James Jack" "von Earl Jones")` and e.g. `(author-name "James Jack von Earl" "Jones")` from being caught by the tests?

jsmaniac:
Ah, yes, the constructor for `author-name` mashes the first, last and suffix together, so the distinction made between the different parts is not actually caught by the tests as long as the appended result is the same. I believe we should make the distinction, though, as hopefully in the future `author-name` could keep this information in extra fields (e.g. to allow the definition of the equivalent of BibTeX styles).

dvanhorn:
I didn't notice #92 -- sorry about that.  I honestly don't care which PR is taken so long as the basic functionality is added.  This patch was just a quick fix for a deadline.

bennn:
closed by #92

PR #116, opened 2017-05-30
acmart: remove subtitle extraction

bennn:
cc @LeifAndersen 

Change implementation of `@title` to just use `\title`.
Do not try to extract a subtitle from the argument.

This fixes a bug, where the `Title` meta-data in the PDF for a document titled
 "Foo" was "oo" (same title with first letter missing). So, e.g., Google chrome
 would show "oo" as the tab title for the document.

- - -

Why remove the `\let` altogether?
  1. I don't think we need it in `scribble/acmart`, because there's a separate `@subtitle` command
  2. Keeping it generates two warnings when running LaTeX:

     ```
       Package hyperref Warning: Token not allowed in a PDF string (PDFDocEncoding):
       (hyperref)                removing `\<let>-command' on input line 674.

       Package hyperref Warning: Token not allowed in a PDF string (PDFDocEncoding):
       (hyperref)                removing `\SSubtitle' on input line 674.
     ```

The "visual" bug was probably due to this warning; after removing the `\let`
 and `\SSubtitle` then only `\SubtitleDrop` and the actual title are left.
I guess `\SSubtitleDrop` removed the first character of the title.
(Why only the first? I don't know.)

The LaTeX warning from `hyperref` is probably because acmart expects the
 argument to `\title` to be a raw string. Though, the docs don't explicitly
 say this. They only say "It is expected that this command [`\title`] is inserted
 by the author of the manuscript."

### DIFF
--- a/scribble-lib/scribble/acmart/style.tex
+++ b/scribble-lib/scribble/acmart/style.tex
@@ -1,12 +1,4 @@
-
-% Define \SXtitle to lift \SSubtitle out:
-\newcommand{\SXtitle}[2][]{\title#1{\let\SSubtitle\SSubtitleDrop#2}\SExtractSubtitle#2\SExtractSubtitleDone}
-\def\SSubtitleDrop#1{}
-\def\SExtractSubtitleDone {}
-\def\SExtractSubtitle{\futurelet\next\SExtractSubtitleX}
-\def\SExtractSubtitleX#1{\ifx{#1}\SSubtitle \let\Snext\SWithSubtitle \else \let\Snext\SExtractSubtitleY \fi \Snext}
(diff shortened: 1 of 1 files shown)


bennn:
I've tested this on 1 document that does not have a subtitle. It looks good.

bennn:
Commit 5e5f96bbaa4c8b1ed445f328ca4c1d977612fe6b did not look good with a `#:short` title.

I tried changing `\SXtitle` to check if the optional argument is supplied, and also changing its callers to just do `[arg]` instead of `[[arg]]`. That looked fine, see commit 7dd19216f0921dcae57a0d4f08a01f12f5fdbdb0 .

But I think we should just use `\title` directly. Commit b50039a02fbb2e7e72217361ff23fa200f3343c6 is the one I propose merging.

LeifAndersen:
Honestly this is fine with me. Feel free to merge.

PR #123, opened 2017-06-23
Small improvements for scriblib/bibtex and acmart

dvanhorn:
- Add support for PhD bib entries in scriblib/bibtex
- Use exact chars in acmart badge commands to avoid escaping URL characters


samth:
This looks good to me.

PR #128, opened 2017-07-11
swap latex rendering of phi and varphi

stchang:
I think `phi` and `varphi` are flipped? eg,

```
#lang scribble/base
ϕφ
```

renders (pdf) as: φϕ


bennn:
yes they are flipped

```
\documentclass{article}
\begin{document}
\[\mbox{\tt phi} = \phi\]

\[\mbox{\tt varphi} = \varphi\]
\end{document}
```

screen shot 2017-07-11 at 13 31 03

rfindler:
Maybe worth checking \varpi and \pi too? I recall trouble with all four of those in the past.

stchang:
The other `varX` letters (`varpi`, `varepsilon`, `varrho`, and `varsigma`) seem consistent with their latex rendering.

rfindler:
Thanks!

PR #129, opened 2017-07-11
autobib book-location: fix capitalization of ordinal editions

stchang:
Fixes capitalization when `book-location` edition is an ordinal.

Eg, the following file:
```
#lang scribble/base
@(require scriblib/autobib)
@(define htdp
  (make-bib #:title "HTDP"
            #:author "MRMS"
            #:is-book? #t
            #:location (book-location #:edition "1st" #:publisher "MIT")
            #:date "2001"))
@(define-cite cite citet gen-bib)
@cite[htdp]
@(gen-bib)
```
capitalizes the edition incorrectly:
```
(MRMS 2001)
Bibliography
MRMS. HTDP. 1St edition. MIT, 2001.
```

### DIFF
--- a/scribble-lib/scriblib/autobib.rkt
+++ b/scribble-lib/scriblib/autobib.rkt
@@ -522,11 +522,17 @@
                 s)])
     s))
 
+(define (string-capitalize str)
+  (if (non-empty-string? str)
+      (let ([chars (string->list str)])
+        (list->string (cons (char-upcase (car chars)) (cdr chars))))
+      str))
+  
 (define (book-location
          #:edition [edition #f]
          #:publisher [publisher #f])
   (let* ([s (if edition
-                @elem{@(string-titlecase (to-string edition)) edition}
+                @elem{@(string-capitalize (to-string edition)) edition}
                 #f)]
          [s (if publisher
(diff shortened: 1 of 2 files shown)


bennn:
I think this is a bug in `string-titlecase`... but I also disagree with that function's documentation.

It says:

> Like string-upcase, but the titlecase conversion only for the first character in each sequence of cased characters in str (ignoring case-ignorable characters).

But I think `string-titlecase` should instead:

> ... capitalize the first character in each sequence of characters in str, unless the first character is case-ignorable.

stchang:
> I think this is a bug in `string-titlecase`

I thought so too at first. For example, it is not consistent with "AP-Style" (http://titlecase.com/)

But then I realized there may just be different standards. For example, titlecasing in C# [1] and python [2] behave like Racket.

So ultimately I decided it was less drastic to change `book-location`.

[1] https://stackoverflow.com/questions/25168617/totitlecase-to-ignore-ordinals-in-c-sharp
[2] https://stackoverflow.com/questions/27080260/how-can-i-change-the-case-of-abbreviated-ordinals-to-lower-while-keeping-the-res

PR #140, opened 2017-09-08
acmart fixes

mflatt:
The first commit makes the acmart style file disable the default Scribble redefinition of `\rmdefault`, since that plays badly with acmart. I'm pretty sure @LeifAndersen had the same problem, but I don't know how she fixed it.

The second one make `grantsponsor` and `grantnum` work. I think the old versions may have been a cut-and-paste error that kept too much from a related function.

The third commit is in the same vein, but for some other functions and more a question of generalizing contracts than making things work at all.

### DIFF
--- a/scribble-doc/scribblings/scribble/acmart.scrbl
+++ b/scribble-doc/scribblings/scribble/acmart.scrbl
@@ -239,28 +239,28 @@ defaults to @racket{Received} for the first occurrence and
   @received[#:stage "accepted"]{June 2009}
 }|}
 
-@defproc[(teaserfigure [content pre-content?] ...) block?]{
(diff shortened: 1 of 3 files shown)


mflatt:
@LeifAndersen Ok with you if I merge this along with an "acmart.cls" upgrade?

LeifAndersen:
Looking more closely, but this looks pretty good to me.

Honestly, I never found a good fix. Since we were down to the wire, I just  manually removed the line from the generated tex file. So having an automated fix is great to me.

LeifAndersen:
Basically, I'm saying:

LGTM

mflatt:
Thanks!

PR #142, opened 2017-09-15
add '\phantomsection' to starx macros

### LINKED ISSUE #124: scriblib/autobib citations should link to bib entry.
When using bibtex+hyperref, clicking on a paper reference links to its location in the bibliography. We should do the same thing with scriblib/autobib.

bennn:
WARNING: I've only tested this on small examples, need to check more about whether this is safe.
In particular, I don't know what this does for documents that don't `\usepackage{hyperref}`.

- - -

Add `\phantomsection` before every call to `\addtocentry` in a Scribble "starx" macro.

This adds a link that hyperref can jump to,
 so the labels added by `scriblib/autobib` work.

Closes #124


bennn:
This fails in documents that don't load `hyperref`. Try again.

PR #144, opened 2017-09-22
autobib: link citations to the bibliography

### LINKED ISSUE #124: scriblib/autobib citations should link to bib entry.

bennn:
1. Add dummy definitions for `hyperref` macros, in case the document doesn't `\usepackage{hyperref}`
2. Add `\phantomsection` in front of table-of-contents entries, so `hyperref` can link to them
3. Change `scriblib/autobib` to insert links & references in its 2 citation styles

For `scribble/manual` and `scribble/acmart` documents, this PR makes citations that link to the bibliography.
And if I don't `\usepackage{hyperref}`, the PDF looks the same but the links are gone.

Closes #124


samth:
Should there be an in-scribble way to specify that you want `hyperref`?

bennn:
On second thought, I should remove the `\providecommand`s because the Scribble LaTeX defaults load `hyperref`. (So those `providecommand`s should go in your alternate `*-load.tex` if you don't want hyperref)

@samth I guess we're missing an in-scribble way to disable `hyperref`. So we should probably (1) document the default packages set in `scribble-load.tex` and (2) make it easy to disable these packages.

mflatt:
#145 also covers this change while trying to address a larger set of problems. That PR is not obviously better, though.

bennn:
A comment after reading #145 : that and this are trying to solve (at least) 3 problems between themselves:

1. Issue #124, that `scriblib/autobib` should make hyperlinks
2. Those hyperlinks should look the same as links generated by LaTeX papers ("Section [N]()" rather than "[Section N]()"; #145 addresses this)
3. All bibliography entries should work as hyperlinks (this PR addresses this with the `\phantomsection`s)

PR #145, opened 2017-10-01
hyperlink customization, especially for acmart

mflatt:
These commits are about making `#lang scribble/acmart` do more of the right things to match the current acmart style – especially with respect to hyperlinks in PDF and link coloring.

I'm not too happy with all the ways that built-in things have to be adjusted here and the way that some things remain hardwired, but this is the best approach I've found so far.

### DIFF
--- a/scribble-doc/scribblings/scribble/base.scrbl
+++ b/scribble-doc/scribblings/scribble/base.scrbl
@@ -613,7 +613,8 @@ Generates a literal hyperlinked URL.}
 @defproc[(secref [tag string?]
                  [#:doc module-path (or/c module-path? #f) #f]
                  [#:tag-prefixes prefixes (or/c (listof string?) #f) #f]
-                 [#:underline? underline? any/c #t])
+                 [#:underline? underline? any/c #t]
+                 [#:link-render-style ref-style (or/c link-render-style? #f)])
          element?]{
 
 Inserts a reference to the section tagged @racket[tag].
@@ -634,32 +635,35 @@ reach the @racket[tag] section. When @racket[#:doc] is not provided,
 the @racket[prefixes] path is relative to any enclosing section (i.e.,
 the youngest ancestor that produces a match).
 
-For HTML output, the generated reference is the hyperlinked title of
-the elements in the section's title content, except that elements with
-the @racket['aux] @tech{style property} are omitted in the hyperlink
-label. If @racket[underline?] is @racket[#f], then the hyperlink is
-rendered in HTML without an underline.
+For the result @racket[link-element], if @racket[ref-style] is not
+@racket[#f], then it is attached as a @tech{style property} and
+affects the rendering of the link. Alternatively, an enclosing
+@racket[part] can have a link-render style that adjusts the rendering
+style for all links within the @tech{part}. See @racket[link-element]
+for more information about the rendering of section references.
 
-For Latex output, the generated reference's format depends on the
-document style. By default, only the section number is shown in the
-reference, but the @racketmodname[scribble/manual] style shows the
-title after the section number. Customize the output (see
-@secref["config"]) by redefining the @ltx{BookRef}, @|etc|, macros (see
-@secref["builtin-latex"]).
-
+If @racket[underline?] is @racket[#f], then a @tech{style} is attached
+to the result @racket[link-element] so that the hyperlink is rendered
+in HTML without an underline
+       
 In Racket documentation that is rendered to HTML, clicking on a
 section title normally shows the @racket[secref] call that is needed
-to link to the section.}
+to link to the section.
+
+@history[#:changed "1.25" @elem{Added the @racket[#:link-render-style] argument.}]}
 
 
 @defproc[(Secref [tag string?]
                  [#:doc module-path (or/c module-path? #f) #f]
                  [#:tag-prefixes prefixes (or/c (listof string?) #f) #f]
-                 [#:underline? underline? any/c #t])
+                 [#:underline? underline? any/c #t]
+                 [#:link-render-style ref-style (or/c link-render-style? #f)])
          element?]{
 
 Like @racket[secref], but if the rendered form of the reference starts
-with a word (e.g., ``section''), then the word is capitalized.}
+with a word (e.g., ``section''), then the word is capitalized.
+
+@history[#:changed "1.25" @elem{Added the @racket[#:link-render-style] argument.}]}
 
 
 @defproc[(seclink [tag string?] 
--- a/scribble-doc/scribblings/scribble/core.scrbl
+++ b/scribble-doc/scribblings/scribble/core.scrbl
@@ -9,6 +9,7 @@
                      scriblib/render-cond
                      xml/xexpr
                      net/url-structs
+                     scriblib/figure
                      (only-in scribble/html-render render-mixin)))
 
 @title[#:tag "core"]{Structures And Processing}
@@ -514,6 +515,11 @@ The recognized @tech{style properties} are as follows:
         module path plus a section-tag string, so that the user can
         create a reference to the section.}
 
+ @item{@racket[link-render-style] structure --- Determines the default
+       rendering of links to sections or other destinations within the
+       section. See also @racket[link-element] and
+       @racket[current-link-render-style].}
+
 ]
 
 The @racket[to-collect] field contains @techlink{content} that is
@@ -525,7 +531,8 @@ sub-parts).
 
 The @racket[parts] field contains sub-parts.
 
-@history[#:changed "1.25" @elem{Added @racket['no-index] support.}]}
+@history[#:changed "1.25" @elem{Added @racket['no-index] support.}
+         #:changed "1.26" @elem{Added @racket[link-render-style] support.}]}
 
 
 @defstruct[paragraph ([style style?] [content content?])]{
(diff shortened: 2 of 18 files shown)


bennn:
Did you try just overriding `ChapRefLocal` etc. in `acmart.tex`?

- - -

If I'm understanding, it looks like the changes to the built-ins are to give a choice between rendering section refs --- or figure refs --- as "Section [N]()" or "[Section N]()". 

It also looks like macros in the LaTeX backend get the section number "N" as one of their arguments. I'm thinking the acmart macros can do this rendering.

- - -

Also fwiw, I don't like the `link-render-style` API. Instead of a `link-render-style`, as a user I'd rather give a function "from section numbers to elements". E.g. if you give me a "4", I'll give you "§ 4". (But instead of a string it'd be an element that uses a style property like `'aux`to decide whether the "§" should be clickable.)

mflatt:
Thanks for taking a look!

Redefining `ChapRefLocal`, etc., works for section links, but that's because section links are constructed by the renderer, at which point it can get the Latex-level tag and pass it to the macro.

The figure library, in contrast, should use `element-target` and `element-link` to wire up hyperlinks, but then it doesn't see the Latex-level key itself and get to pass it off to a Latex macro. (The older approach via "private/counter" duplicates the management of Latex-level tags, which is bad.)

Along those lines, the `link-render-style` API is intended to work for other kinds of links besides section links, such as figure links, so that's why it doesn't defer the choice of how to spell "section" to the render style.

Finally, the intent was also to provide some control over HTML output, so I was looking for something more at the Scribble level than the Latex-macro level (although I experimented with both).

bennn:
Ok thanks for clarifying. I hadn't thought about issues with the figure library --- and I didn't know about the older approach.

I'll read this again & give lower-level comments later today.

scribble-doc/scribblings/scribble/core.scrbl
```diff
@@ -1421,7 +1444,32 @@ number for each subsection. If @racket[numberer-step] produces a plain
...
 subsection numbers. See also @racket[collected-info].
 
-@history[#:added "6.4"]}
+@history[#:added "1.1"]}
+
+
+@defstruct[link-render-style ([mode (or/c 'default number)])]{
+
+Used as a @tech{style property} for a @racket[part] or a specific
+@racket[link-element] to control the way that the link is rendered.
```
bennn: Replace "the way the link is rendered" with "what parts of the `link-element` are hyperlinked" ?

scribble-doc/scribblings/scribble/core.scrbl
```diff
@@ -1421,7 +1444,32 @@ number for each subsection. If @racket[numberer-step] produces a plain
 string for the rendered number, then it is not added as a prefix to
 subsection numbers. See also @racket[collected-info].
 
-@history[#:added "6.4"]}
+@history[#:added "1.1"]}
+
+
+@defstruct[link-render-style ([mode (or/c 'default number)])]{
```
bennn: typo: `number` => `'number`

About naming: I'm thinking `'default` should be `'all`, so it's phrased as an answer to the question "what parts of the `linke-element` are clickable?"

scribble-doc/scribblings/scribble/core.scrbl
```diff
@@ -1421,7 +1444,32 @@ number for each subsection. If @racket[numberer-step] produces a plain
...
+@margin-note{The @racket[figure-ref] and @racket[Figure-ref] functions
+             from @racketmodname[scriblib/figure] also adapt to
+             a @racket[link-render-style] mode.}
+
+@history[#:added "1.25"]}
+
+
+@defparam[current-link-render-style style link-render-style?]{
+
+A parameter that determines the default rendering style for a section
```
bennn: Replace "section link" with `link-element` ? (Maybe there should be a `@tech` for this)

scribble-doc/scriblib/scribblings/figure.scrbl
```diff
@@ -85,14 +85,28 @@ The @racket[left] binding is a synonym for @racket[left-figure-style],
...
 
 
-@defproc[(figure-ref [tag string?] ...+) element?]{
+@defproc[(figure-ref [tag string?] ...+
+                     [#:link-render-style link-style (or/c link-render-style? #f)])
+         element?]{
+
+Generates a reference to one or more figures, using a lowercase word ``figure''.
+
+If @racket[link-style] or (at the time of rendering)
```
bennn: Is this "at the time of rendering" caveat the reason `secref` etc. don't use `(current-link-render-style)` as a default argument?

scribble-doc/scriblib/scribblings/figure.scrbl
```diff
@@ -85,14 +85,28 @@ The @racket[left] binding is a synonym for @racket[left-figure-style],
...
+@racket[(current-link-render-style)] indicates the @racket['number]
+style mode, then the word ``figure'' itself is not hyperlinked;
+otherwise, the word @racket[fgure] is hyperlinked together with the
```
bennn: typo: `@racket[fgure]` => `"figure"` ?

scribble-lib/scribble/acmart/acmart.tex
```diff
@@ -23,4 +23,6 @@
 % Normally gets re-written by the title macro:
 \newcommand{\SSubtitle}[1]{{\bf #1}}
 
-
+% Use ACM color; it would be better to use `citecolor` here somehow,
+% but I can't figure out how to do that
+\newcommand{\AutobibLink}[1]{\color{ACMPurple}{#1}}
```
bennn: Can this take 2 arguments and call `\hyperref` instead of `\color` ?

scribble-lib/scribble/acmart/lang.rkt
```diff
@@ -142,8 +142,16 @@
...
+                ;; Ensure that "acmart.tex" is used, since "style.tex"
+                ;; re-defines commands.
+                (cons invisible-element-to-collect-for-acmart-extras
+                      (part-to-collect doc))]
+               [style (let ([s (part-style doc)])
+                        (struct-copy style s
+                                     [properties
+                                      ;; Immitate Latex-based links where only the
+                                      ;; number part is hyperlinked.
+                                      (cons (link-render-style 'number)
```
bennn: Should `latex-render.rkt` set this default instead?

scribble-lib/scribble/core.rkt
```diff
@@ -162,6 +162,36 @@
...
+  #:property
+  prop:serializable
+  (make-serialize-info
+   (lambda (s)
+     (vector (link-render-style-mode s)))
+   #'deserialize-link-render-style
+   #f
+   (or (current-load-relative-directory) (current-directory))))
+
+(provide deserialize-link-render-style)
```
bennn: Why not put this provide with the others?

scribble-lib/scribble/latex-render.rkt
```diff
@@ -21,6 +21,7 @@
 (define disable-images (make-parameter #f))
 (define escape-brackets (make-parameter #f))
 (define suppress-newline-content (make-parameter #f))
+(define disable-hyperref (make-parameter #f))
```
bennn: I'm not sure what the goal of this parameter is.

If it's trying to decide whether to print `\hyperref{....}` or just the contents `....`, I think it should just always do `hyperref` (or a Scribble macro)

scribble-lib/scribble/core.rkt
```diff
@@ -162,6 +162,36 @@
...
+                         (lambda (tag init-val)
+                           (error "cannot allocate link-render-style for cycle"))))
+
+(define current-link-render-style (make-parameter (link-render-style 'span)))
+
+(provide
+ link-render-style?
+ link-render-style-mode
+ (contract-out
+  [link-render-style ((or/c 'span 'number)
```
bennn: typo: `'span` => `'default`
(or the docs are wrong)

scribble-lib/scribble/base.rkt
```diff
@@ -549,45 +549,55 @@
...
   (make-target-element #f (decode-content body) `(elem ,t)))
 (define (elemref #:underline? [u? #t] t . body)
   (make-link-element (if u? #f "plainlink") (decode-content body) `(elem ,t)))
 
-(define (secref s #:underline? [u? #t] #:doc [doc #f] #:tag-prefixes [prefix #f])
-  (make-link-element (if u? #f "plainlink") null (make-section-tag s #:doc doc #:tag-prefixes prefix)))
-(define (Secref s #:underline? [u? #t] #:doc [doc #f] #:tag-prefixes [prefix #f])
+(define (secref s #:underline? [u? #t] #:doc [doc #f] #:tag-prefixes [prefix #f]
+                #:link-render-style [link-style #f])
+  (make-link-element (let ([name (if u? #f "plainlink")])
```
bennn: Does this mean LaTeX section refs render `"section N"` and not `"section~N"` ?

scribble-lib/scriblib/autobib.tex
```diff
@@ -8,3 +8,7 @@
 \newcommand{\Autocolbibentry}[1]{\setlength{\ABcollength}{\linewidth-5ex}\parbox[t]{\ABcollength}{#1\vspace{1.0ex}}}
 
 \newcommand{\Autobibref}[1]{#1}
+
+\makeatletter
+\@ifundefined{AutobibLink}{\newcommand{\AutobibLink}[1]{#1}}{}
+\makeatother
```
bennn: Could this be `\providecommand{\AutobibLink}....` ?

scribble-doc/scribblings/scribble/base.scrbl
```diff
@@ -613,7 +613,8 @@ Generates a literal hyperlinked URL.}
 @defproc[(secref [tag string?]
                  [#:doc module-path (or/c module-path? #f) #f]
                  [#:tag-prefixes prefixes (or/c (listof string?) #f) #f]
-                 [#:underline? underline? any/c #t])
+                 [#:underline? underline? any/c #t]
+                 [#:link-render-style link-style (or/c link-render-style? #f)])
```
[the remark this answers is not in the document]
mflatt: The intent is that `current-link-render-style` is a render-time parameter, as opposed to a document-construction time parameter. That way, a section can "inherit" any specific setting for an enclosing section.

scribble-doc/scriblib/scribblings/figure.scrbl
mflatt: Yes

scribble-lib/scribble/acmart/acmart.tex
mflatt: Not in the current design, because the `\hyperref` part is triggered by `link-element`ness, while `\AutobibLink` comes from the style.

scribble-lib/scribble/acmart/lang.rkt
mflatt: I don't know what you have in mind here. The intent is to make acmart render differently than other existing Scribble->Latex documents.

scribble-lib/scribble/latex-render.rkt
mflatt: Putting a `\hyperref` in things like `\section` causes the table of contents to go bad (and then Latex errors out), so the internal parameter is meant to work around problems like that.

scribble-lib/scriblib/autobib.tex
mflatt: Apparently so. (I had never heard of `\providecommand`.)

mflatt:
Revised, especially the docs, based on @bennn's comments.

scribble-lib/scribble/core.rkt
mflatt: The `provide`s in this file are grouped more with the implementations, and the `deserialize-....` exports are consistently (I think) grouped with the deserializers.

Really, this file should be split into several, and the deserializers should be in submodules. But I didn't tackle that refactoring here.

scribble-lib/scribble/acmart/acmart.tex
bennn: Ok.
I also couldn't find a way to use `citecolor`

scribble-lib/scribble/acmart/lang.rkt
bennn: Sorry, it wasn't a useful thought.
What I had in mind was "well since this is here to imitate LaTeX links, maybe all scribble->LaTeX documents should use the `'number` style"

scribble-doc/scribblings/scribble/core.scrbl
```diff
@@ -982,18 +989,65 @@ field.}
 
 @defstruct[(link-element element) ([tag tag?])]{
 
-Hyperlinks the content to @racket[_tag].
+Represents a hyperlinks to @racket[_tag].
```
bennn: typo: `hyperlinks`

PR #148, opened 2017-10-12
scribble.tex: add phantomsection to TOC links

bennn:
add `\phantomsection`s to make valid hyperref targets

(just like #142 )


bennn:
Here's a small test for this:  [acmsmall-vs-sigplan.tar.gz](https://github.com/racket/scribble/files/1715054/acmsmall-vs-sigplan.tar.gz)

If you unzip & run `make`, you'll get two pdfs: `acmsmall.pdf` and `sigplan.pdf`. Both have a citation near the top of their first page.

- Without this commit, clicking the citation in `acmsmall.pdf` does nothing. Clicking in `sigplan.pdf` jumps to the bibliography.
- With this commit, clicking in either document jumps to the bibliography.

PR #153, opened 2017-11-26
updated CSS for documentation

mbutterick:
PS `manual-fonts.css` is generated from a Pollen file and a directory of fonts. The `scribble` repo doesn’t seem like the right place to store those (since that would make a `scribble` build dependent on Pollen). 

I see two options, though I don’t know which would fit better with Racket build policies:

1) Put them in `https://github.com/racket/visual-design` (so at least there’s a record of how they were made)

2) Convert the Pollen source to a `scribble/text` source file, and make it part of the build process.


mbutterick:
This is the CSS currently demonstrated on http://mbutterick.github.io/racket-doc-redo/doc/

Only a few cases of brain damage were reported. They have been fixed.

PR #156, opened 2018-01-02
Provide special?

soegaard:
For programs using scribble/html/xml to represent html, the predicate used to represent entities were missing.
The contract outputable can not be used instead since it has been effectively be removed.


bennn:
resolved by #165

soegaard:
Oh! I forgot about this issue. Thanks for closing it.

Issue #159, opened 2018-01-29
scribble/acmart doesn't provide include-abstract


dfeltey:
The scribble/acmart documentation claims the language provides the `include-abstract` function, but it seems to not actually be provided or defined.

LeifAndersen:
Thanks for pointing this out @dfeltey . As @bennn said, #160 closes this. So I'll move any discussion to there.

PR #160, opened 2018-01-31
Add include-abstract.

LeifAndersen:
The docs claimed it was there, but we missed it when creating
scribble/acmart. As such, this commit adds it in.

### DIFF
--- a/scribble-lib/scribble/acmart.rkt
+++ b/scribble-lib/scribble/acmart.rkt
@@ -86,7 +86,8 @@
   (->* () () #:rest (listof pre-content?)
        any/c)])
 (provide
-  invisible-element-to-collect-for-acmart-extras)
+  invisible-element-to-collect-for-acmart-extras
+  include-abstract)
 
 (define-syntax-rule (defopts name ...)
   (begin (define-syntax (name stx)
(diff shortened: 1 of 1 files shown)


bennn:
Closes #159

scribble-lib/scribble/acmart.rkt
```diff
@@ -159,6 +160,20 @@
...
+  (unless (null? (part-parts p))
+    (error 'include-abstract "abstract part has sub-parts: ~e" (part-parts p)))
+  (when (part-title-content p)
+    (error 'include-abstract "abstract part has title content: ~e" (part-title-content p)))
+  (part-blocks p))
+
+(define-syntax-rule (include-abstract mp)
+  (begin
+    (require (only-in mp [doc abstract-doc]))
+    (make-nested-flow abstract-style (extract-abstract abstract-doc))))
```
bennn: Can this call the `abstract`function?

LeifAndersen: Sadly no. Because we don't want it to re-sanitize the string. We 'could' add a secret extra parameter to do that, but...meh?

LeifAndersen:
Okay cool. I'll go ahead and merge this then.

PR #161, opened 2018-02-04
defthing: use #:id expression instead of 'id'

bennn:
Change implementation of `defthing` so that if the user gives an `id-expr` via the `#:id` keyword, then `defthing` never uses the default `id`.

I think this is right, based on the docs. But maybe the docs should change.

cc @dedbox for reporting this

- - -

Example program:

```
#lang scribble/manual

@defthing[#:id (datum->syntax #f 'foo) dont-use integer?]{
  Something
}
```

Before this PR:
screen shot 2018-02-04 at 16 50 08

After this PR:
screen shot 2018-02-04 at 16 50 18

### DIFF
--- a/scribble-lib/scribble/private/manual-proc.rkt
+++ b/scribble-lib/scribble/private/manual-proc.rkt
@@ -1039,12 +1039,13 @@
      #'(with-togetherable-racket-variables
         ()
         ()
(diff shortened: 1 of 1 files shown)


bennn:
e2ea830361138212974c3637e9e928bc1f753bde

PR #166, opened 2018-03-25
lncs: add basic .css file

bennn:
Copy `scribble/sigplan`'s style file to `scribble/lncs` because the HTML renderer expects to find an `lncs.css` file.

Closes https://github.com/racket/racket/issues/2010


bennn:
d4d38e1ac4f9f2fc98ce7f8d658e5c31aa349fed

PR #168, opened 2018-03-29
acmart: avoid mathabx '\bigtimes'

bennn:
The [`acmart`](https://www.acm.org/binaries/content/assets/publications/consolidated-tex-template/acmart.pdf) LaTeX class uses the `newtxmath` package, if available. The `newtxmath` package provides a `\bigtimes` command.

Scribble's `scribble-load.tex` uses the `mathabx` package by default, which also provides a `\bigtimes` command. This means that `scribble/acmart` documents don't work on systems with `newtxmath` installed. Example:

```
#lang scribble/acmart
@title{Title}
@author{Author}
;; raco scribble --pdf file.scrbl ==> BOOM!
```

Since `scribble/acmart` documents should use the symbols that `acmart` provides, this pull request hides the `mathabx` version of `\bigtimes`

Thank you @pnwamk for reporting and thank you @LeifAndersen for helping debug

TODO:
- [X] test on TexLive 2017 (with `newtxmath`)
- [x] text on TexLive 2015 (without `newtxmath`)

### DIFF
--- a/scribble-lib/scribble/acmart/acmart-load.tex
+++ b/scribble-lib/scribble/acmart/acmart-load.tex
@@ -3,6 +3,15 @@
 % Avoid package option conflict
 \renewcommand\packageColor\relax
 \renewcommand\packageTocstyle\relax
+\renewcommand\packageMathabx{
+  \ifx\bigtimes\undefined
+    \usepackage{mathabx}
+  \else
+    \let\newtxmathbigtimes=\bigtimes
+    \let\bigtimes\undefined
+    \usepackage{mathabx}
+    \let\bigtimes=\newtxmathbigtimes
+  \fi}
 \let\Footnote\undefined
 \let\captionwidth\undefined
 \renewcommand{\renewrmdefault}{}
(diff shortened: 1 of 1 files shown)


bennn:
On second thought, `scribble/acmart` should probably not import `mathabx` by default because it changes a lot of symbols [(doc)](https://ctan.org/pkg/mathabx?lang=en). I'm not sure if these changes conflict with the ACM's styles, but it seems safer to let Scribble users opt-in to `mathabx`.

The current version of this PR keeps `mathabx` on "old" systems (without `newtxmath`) and does not import `mathabx` on newer systems.

rfindler:
This kind of change seems likely to cause old papers not to build anymore, something I think should be avoided if at all possible. Can you please find another way?

bennn:
I don't think this can affect old papers.
- If `\bigtimes` is undefined, we still require `mathabx` like before.
- If `\bigtimes` is defined, then the paper couldn't have built with `mathabx` before, so it's safe to not require it anymore

rfindler:
oic. Thanks!

bennn:
6e00000bb5a1db1f1cf1827ab9710b4eced8f1a8

PR #171, opened 2018-04-15
scribble/bnf: Add contracts & correct docs

### LINKED ISSUE #170: Contract issues in scribble/bnf
I'm opening an issue to note some problems with contracts in `scribble/bnf`. First of all, none of the exports of the module enforce their contracts. Adding them, though, will be more difficult because the contracts in the [documentation](http://docs.racket-lang.org/scribble/bnf.html) are incorrect.

I discovered the lack of contracts when I got a very ugly error message from the guts of `BNF`. (I'd passed it a deeply nested quasiquoted list, and I'd left out a `,` somewhere.) When I initially tried to make a wrapper with `define/contract`, I then discovered that `(BNF-seq)` produces `""`, which is not an `element?`, even though the documentation both says that `BNF-seq` will produce an `element?` _and_ says that the arguments to BNF must be `(cons element? (listof (or/c block? element?)))`, though obviously `(BNF (cons (nonterm "example") (list (BNF-seq))))` actually works just fine.

If I get a chance, I'll try to figure out what the contracts for these functions actually should be and update both the code and the docs accordingly.

LiberalArtist:
Closes #170

### DIFF
--- a/scribble-doc/scribblings/scribble/bnf.scrbl
+++ b/scribble-doc/scribblings/scribble/bnf.scrbl
@@ -1,6 +1,7 @@
 #lang scribble/doc
 @(require scribble/manual "utils.rkt" scribble/bnf
-          (for-label scribble/bnf))
+          ;; "utils.rkt" provides scribble/bnf for-label
+          )
 
 @title[#:tag "bnf"]{BNF Grammars}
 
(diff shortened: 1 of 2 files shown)


LiberalArtist:
I'm not sure if there's time at this point before the release, but I wanted to follow up about this. @samth, you're the only one who shows up in the Git history as having touched this file.

samth:
That's an artifact of my having moved the file, @mflatt is the only person with commits that touch the current content.

PR #173, opened 2018-05-03
Parameterize figure captions by language.

LeifAndersen:
acmart uses a different font and figure prefix for its captions than other latex styles. This allows different #langs to parameterize various properties about figures (defaulting to the existing ones), so that our figure library can match acmart when using `#lang scribble/acmart`.


LeifAndersen:
I have updated the version of acmart to the latest latex class (1.53). If you don't want to merge those changes with this PR we can always just do a cherry pick.

Also tagging @bennn and @pnwamk.

scribble-lib/scribble/private/lang-params.rkt
```diff
@@ -0,0 +1,12 @@
...
+
+(provide (all-defined-out))
+
+;; Some latex formats have different requirements on how
+;;   figures, citations, etc. are displayed. This allows different
+;;   scribble langs to handle them.
+
+(define default-label-text (make-parameter "Figure"))
+(define default-label-sep (make-parameter ": "))
+(define default-caption-style (make-parameter #f))
```
bennn: Can you rename these first 3 to `default-figure-....`

bennn: Also, add a comment describing the type of each parameter

LeifAndersen: Agreed, done.

scribble-lib/scribble/acmart.rkt
```diff
@@ -9,6 +9,7 @@
          scribble/html-properties
          scribble/latex-properties
          scribble/private/tag
+         scribble/private/lang-params
```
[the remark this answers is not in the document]
LeifAndersen: I'd be happy to change that.

scribble-lib/scribble/acmart/lang.rkt
```diff
@@ -23,22 +24,24 @@
...
-	   [((review #f) . body)
-	    (set! review? "review=false")
-	    (loop #'body)]
+           [((review #t) . body)
+            (set! review? "review=true")
+            (loop #'body)]
+           [((review #f) . body)
+            (set! review? "review=false")
+            (loop #'body)]
 	   [(review . body)
```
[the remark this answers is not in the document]
LeifAndersen: Mmm...looks looks wrong to me. It's probably because the original file used tabs instead of spaces...which DrRacket and github probably treat differently...

PR #174, opened 2018-05-15
Fixes #120: margin notes can collide with each other

### LINKED ISSUE #120: margin notes can collide with each other
When two margin notes are defined close to each other, they can overlap in the HTML version (in the example below, `Hello` and `world` are typeset above one another).

```
#lang scribble/manual
@(require scriblib/footnote)
Foo@note{Hello} bar@note{world}.
```

On chromium, I can fix this by changing in https://github.com/racket/scribble/blob/master/scribble-lib/scriblib/footnote.css

```css
.NoteBox {
  …
  height: 0em;
  …
}
```

to

```css
.NoteBox {
  …
  height: auto;
  clear: right;
  …
}
```

Is there any particular reason for having this `height: 0em` in the first place (e.g. compatibility with an older/another browser)?

If this change seems correct, I can submit a PR.

jsmaniac:


jsmaniac:
If there are two notes in the same line of text, they would both "flow right" and the second one would be superimposed on top of the first. In order to have the note also move down to the first available space, you have to indicate that it should have "cleared" all the other "float"ing elements, and place itself below that point.

See here for a demo: https://jsfiddle.net/9dj2wfpt/1/

PR #176, opened 2018-06-14
add optional #:note to 'bib-entry'

bennn:
Add an optional argument to bib-entry to typeset a comment directly
after a citation --- with no punctuation between the end of the citation
and start of the comment.

Example uses:

- making an annotated bibliography, with an element that appears just
  below each citation

- adding "accessed" dates just after a URL, e.g.
  `(bib-entry .... #:url "...." #:note ", accessed 2018-06-14")`

Example document:

```
#lang scribble/manual
@title{Doc}

@(bibliography
   (bib-entry #:key "WT18"
              #:author "J. Wang and S. Z. Taylor"
              #:title "Parabolic Monoids and Concrete Galois Theory"
              #:location "Mathgen"
              #:date "2018"
              #:url "http://thatsmathematics.com/mathgen/"
              #:note ", accessed 2018-06-14")
   (bib-entry #:key "RJ18"
              #:author "D. Raman and R. A. Jones"
              #:title "On the Computation of Equations"
              #:location "Mathgen"
              #:date "2018"
              #:note @elem{@linebreak[]@linebreak[]@emph{Abstract:}
                Let @math{f_Z = U} be arbitrary. In @cite{WT18} the authors
                address the locality of hyper-holomorphic planes under the
                additional assumption that @math{Θ' = -1}. We show that there
                exists an one-to-one, stable, invertible and differentiable
                real, super-conditionally complete subset.}))
```

Output:

screen shot 2018-06-14 at 14 39 39

### DIFF
--- a/scribble-doc/scribblings/scribble/manual.scrbl
+++ b/scribble-doc/scribblings/scribble/manual.scrbl
@@ -1909,7 +1909,8 @@ order as given.}
                     [#:author author (or/c #f pre-content?) #f]
                     [#:location location (or/c #f pre-content?) #f]
                     [#:date date (or/c #f pre-content?) #f] 
-                    [#:url url (or/c #f pre-content?) #f])
+                    [#:url url (or/c #f pre-content?) #f]
+                    [#:note note (or/c #f pre-content?) #f])
          bib-entry?]{
 
 Creates a bibliography entry. The @racket[key] is used to refer to the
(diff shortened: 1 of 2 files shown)


LeifAndersen:
Honestly, accessed seems like it should be its own field. Like:

#:accessed "..."

Because that will very from style to style.

bennn:
Leif is right, putting ", accessed" in a note is a bad way to add that information. I can remove that example from the commit message.

I'm not sure now whether the note should come directly after the citation, or have 1 linebreak in between. (My preference is for putting it directly after.)

PR #180, opened 2018-08-28
add docs for acmart options

jbclements:
Add descriptions of the acmart options such as "review", "anonymous", and "authorversion". Docs taken from the ACM documentation as of today (2018-08-28)

### DIFF
--- a/scribble-doc/scribblings/scribble/acmart.scrbl
+++ b/scribble-doc/scribblings/scribble/acmart.scrbl
@@ -74,6 +74,54 @@ number of options may be used:
 }|
 
 If multiple font size options are used, all but the last are ignored.
+
+The @link["https://www.acm.org/binaries/content/assets/publications/consolidated-tex-template/acmart.pdf"
+ ]{ACM documentation} (version 1.54, 2018-07-16, by Boris
+Veytsman) provides these defaults and descriptions:
+
+@tabular[#:style 'boxed
+         #:sep @hspace[1]
+         ;#:column-properties '(left left left)
+         #:row-properties '(bottom-border ())
+         (list
+          (list @bold{name} @bold{default} @bold{description})
+          (list "review" "false"
+                "A review version: lines are numbered and\
+ hyperlinks are colored")
+          (list "screen" "see text"
+                "A screen version: hyperlinks are colored")
+          (list "natbib" "true"
+                "Whether to use the natbib package")
+          (list "anonymous"
+                "false"
+                "Whether to make author(s) anonymous")
+          (list "authorversion"
+                "false"
+                "Whether to generate a special version\
+ for the authors’ personal use or posting")
+          ;; these options are documented in ACM docs but don't
+          ;; appear to exist in the scribble acmart format:
+          #;(
+          (list "nonacm" "false"
+                "Use the class typesetting options for a non-ACM\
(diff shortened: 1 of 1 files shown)


jbclements:
Good idea! Okay, added link & version number for ACM docs.

jbclements:
I copied all of the options in the ACM docs, but I commented out those that didn't appear in the existing docs. I've verified that all of the ones whose documentation is visible are in fact supported by the `scribble/acmart` language.

But! Another question just occurred to me. For options such as natbib, the default is supposed to be "true". If that's true, though, how is one supposed to indicate in the #lang line that natbib is supposed to be *false*?  I went digging a bit into the source code, and it appears very much to me that in fact, the default values for all of these options is false, with the result that a basic document that begins

``` racket
#lang scribble/acmart @sigplan
```

will not get the correct default for the natbib option. It appears to me that--in order to preserve the notion that `natbib` is default true--the 'natbib' option should probably be replaced with a `no-natbib` option.

As it is, the documentation I've supplied is deeply misleading, because the notion of a "default true" option doesn't make any sense. 

Oh dear.

bennn:
Should we export `natbib=true` and `natbib=false` identifiers?

jbclements:
Okay, more investigation reveals that with @natbib missing, the output latex just omits all mention of natbib--thus defaulting to true. So in fact there's no way in the current system (AFAICT) to actually produce a latex document with `natbib=false`.

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -74,6 +74,47 @@ number of options may be used:
 }|
 
 If multiple font size options are used, all but the last are ignored.
+
+The ACM documentation (version 1.54, 2018-07-16, by Boris
```
bennn: maybe use a `@margin-note{}` instead of parentheses?

jbclements: Oh my... in fact, the "margin-note" function (when used in that file, at least) produces extraordinarily prominent text. I don't think that's an improvement. Here's what it looks like...
screen shot 2018-08-28 at 14 04 16

jbclements:
I don't think I like `natbib=true` and `natbib=false` options, because they're not consistent with the other options. I think I still favor `no-natbib`, and with the revelation that omitting `natbib` doesn't actually turn natbib off, I'm no longer concerned about breaking existing documents, because this change won't change the meaning of existing documents.

bennn:
I was thinking, all the options could use the `=true` and `=false` style where needed.

But `no-natbib` is fine with me.

LeifAndersen:
So, I can't remember if this is this version of acmart, or the version that @fare (https://github.com/fare/acmart) and I were working on (before David pushed his to master...), but you can do @natbib[#f] to disable natbib, and the same for other options with true/false flags.

jbclements:
Ah! works perfectly. Didn't occur to me that these might also be usable as parameter-like functions. Looks like I should document this behavior, as well.

LeifAndersen:
Yes please.

image

LeifAndersen:
Woops...ya, that came out _way_ bigger than I meant...oh well. Anyway...

LGTM

jbclements:
okay, squashed commits, now merging onto racket/scribble

PR #181, opened 2018-09-12
Add an identifier for acmart's nonacm flag.

LeifAndersen:


scribble-lib/scribble/acmart/lang.rkt
```diff
@@ -22,12 +22,13 @@
            [natbib? #f]
            [anonymous? #f]
            [authorversion? #f]
-           [font-size #f])
+           [font-size #f]
+           [nonacm? #f])
```
bennn: Its not a boolean-valued variable, so change the name to `nonacm` ?

LeifAndersen: I would, but in this case I thought it better to follow the convention of the other variables. So maybe we could remove the `?` on all of them?

LeifAndersen:
Okay @bennn I've added the id to the docs.

scribble-lib/scribble/acmart/lang.rkt
bennn: Oh I didn't notice them --- nevermind

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -58,6 +58,7 @@ all mutually exclusive.}
 @defidform[natbib]
 @defidform[anonymous]
 @defidform[authorversion]
+@defidform[noacm]
 @defidform[9pt]
 @defidform[10pt]
 @defidform[11pt]
```
bennn: There's also a `@tabular` of defaults below, copied from Table 2 (page 8) of the acmart pdf. `nonacm` should be in the tabluar.
https://www.acm.org/binaries/content/assets/publications/consolidated-tex-template/acmart.pdf

LeifAndersen:
Okay Ben. I have added it to the docs, and also added the other variables that acmart was missing. If we need any more variables I very much think we need a macro for the macro.

scribble-lib/scribble/acmart/lang.rkt
```diff
@@ -116,7 +156,7 @@
             (loop #'body)]
            
            [body
-            #`(#%module-begin id (post-process #,review? #,screen? #,natbib? #,anonymous? #,authorversion? #,font-size #,format?) () . body)])))]))
+            #`(#%module-begin id (post-process #,review? #,screen? #,natbib? #,anonymous? #,authorversion? #,font-size #,nonacm? #,format?) () . body)])))]))
```
bennn: This line needs `#,timestamp? #,authordraft? #,acmthm?`

PR #182, opened 2018-09-30
Add book-chapter-location to autobib

mlemmer:
Added an option to autobib for a chapter or part of a book or anthology

### DIFF
--- a/scribble-doc/scriblib/scribblings/autobib.scrbl
+++ b/scribble-doc/scriblib/scribblings/autobib.scrbl
@@ -236,6 +236,17 @@ Combines elements to generate an element that is suitable for
 describing a dissertation.}
 
 
+@defproc[(book-chapter-location [title any/c]
+                           [#:pages pages (or (list/c any/c any/c) #f) #f]
+                           [#:section section any/c #f]
+                           [#:volume volume any/c #f]
+                           [#:publisher publisher any/c #f])
+         element?]{
+
+Combines elements to generate an element that is suitable for
+describing a paper's location within a chapter or part of a book.}
+
+
 @defproc[(author-name [first any/c]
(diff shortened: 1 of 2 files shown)


scribble-lib/scriblib/autobib.rkt
```diff
@@ -563,6 +565,27 @@
          #:degree [degree "PhD"])
   @elem{@to-string[degree] dissertation, @to-string[org]})
 
+(define (book-chapter-location
+         location
+         #:pages [pages #f]
+         #:series [series #f]
+         #:volume [volume #f]
+         #:publisher [publisher #f])
+  (let* ([s @elem{In @italic{@elem{@|location|}}}]
```
bennn: The contract on `location` is `any/c`, so I think this should be `@italic{@to-string[location]}`

... maybe the contract should be `string?` because that's what this line is assuming?

scribble-lib/scriblib/autobib.rkt
```diff
@@ -563,6 +565,27 @@
...
+         [s (if series
+                @elem{@|s|, @(format "~a" series)}
```
bennn: ps: `(to-string X)` is a shorter way to write `(format "~a" X)`

scribble-doc/scriblib/scribblings/autobib.scrbl
```diff
@@ -236,6 +236,17 @@ Combines elements to generate an element that is suitable for
 describing a dissertation.}
 
 
+@defproc[(book-chapter-location [title any/c]
+                           [#:pages pages (or (list/c any/c any/c) #f) #f]
```
bennn: could you indent `[#:pages` to line up with `[title` ?

PR #186, opened 2018-12-18
Synchronize the document of #:email in author with its actual the contract

shhyou:
The `#:email` argument in `author` takes `(or/c pre-content? email? (listof email?) #f)` instead of `(or/c pre-content? (listof pre-content?))`. This also seems to be the case in the code. Plus, all `(listof pre-content?)` values are already a `pre-content?`.

### DIFF
--- a/scribble-doc/scribblings/scribble/acmart.scrbl
+++ b/scribble-doc/scribblings/scribble/acmart.scrbl
@@ -161,8 +161,8 @@ Specifies a subtitle.}
                         #f)
                   #f]
                  [#:email email
-                  (or/c pre-content? (listof pre-content?) #f)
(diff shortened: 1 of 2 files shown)


bennn:
Looks good

While we're here, what do you think about removing `#f` from the contract in the docs & implementation? Currently, `author` fails on a `#f` email:

```
Welcome to Racket v7.1.0.10.
> (require scribble/acmart)
> (author #:email #f "a")
; in-list: contract violation
;   expected: list?
;   given: #f
; [,bt for context]
```

shhyou:
Good catch, I totally missed that! `#f` was probably meant for `'()`.

PR #196, opened 2019-02-13
Add block? contract to typeset-code

pmatos:
I think something should be said about what `block?` does but I am not comfortable enough with the scribble code to write it down.


bennn:
https://github.com/racket/scribble/commit/ac37653988764b089e41a9c9fe1c1db5887312f1 adds an explanation

PR #198, opened 2019-02-28
make 'verbatim' suppress line breaks in LaTeX output

bennn:
by re-defining Scribtexttt within a LaTeX-level verbatim environment,
because that command is wrapped around each line of the output

cc @mflatt

context: <https://groups.google.com/d/msg/racket-users/Ylkbre0SdQc/ivgK3e2hAQAJ>

### DIFF
--- a/scribble-lib/scribble/scribble.tex
+++ b/scribble-lib/scribble/scribble.tex
@@ -388,7 +388,7 @@
 \newcommand{\Svcenter}[1]{$\vcenter{#1}$}
 
 % Verbatim
-\newenvironment{SVerbatim}{}{}
+\newenvironment{SVerbatim}{\let\Sscribtextttold\Scribtexttt\renewcommand{\Scribtexttt}[1]{\mbox{\Sscribtextttold{##1}}}}{}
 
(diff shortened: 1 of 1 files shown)


mflatt:
That's a neat trick. My only concern is that a Latex trick like this may interact badly with other Latex tricks.

Would it make more sense to put a style on the Scribble-level `paragraph` that represents one line within `verbatim`, instead of on the whole `verbatim`? Maybe keep `SVerbatim` and add `SVerbatimLine`? I was trying to stay away from a larger change like that, but maybe it would be a better choice than trying to work completely at the Latex level.

bennn:
Yes it makes sense to do this at the Scribble level instead, with a "verbatim line" style.

I guess it also makes sense to leave this alone until someone has a need for it.

p.s. Yesterday I was wondering why `verbatim`'s [`make-line`](https://github.com/racket/scribble/blob/54606422140136dbc8f0759f2d44e757a0193ac4/scribble-lib/scribble/base.rkt#L757) helper function used the `'omitable` style --- does `'omitable` do anything for tex output? A "verbatim line" style probably would have been clearer.

samth:
Should this be closed?

PR #200, opened 2019-03-25
Add https://blog.racket-lang.org/2012/06/submodules.html as an example for in-source documentation

IlnarSelimcan:
I couldn't locate gif.rkt in the `file' collection. Seeing a complete working example was helpful to get started. I think it will be as useful for other Racketeers (especially beginners like myself).

If gif.rkt is indeed gone due to refactorings, I guess the reference to it should be removed.


samth:
That file is here: https://github.com/racket/draw/blob/master/draw-lib/file/gif.rkt but no longer seems to use in-source documentation.

bennn:
Thanks! I will remove the `gif` reference.

PR #202, opened 2019-04-18
search box: change '"s" for search' behavior

bennn:
1. Focus on the search box for either "s" or "S"
   ... accepting only "s" makes sense to me, but the comment said it
   accepted "S" and well why not
2. Look for a "keyup" event instead of key press, so that pressing "s"
   ONLY focuses on the box and does not focus-and-write-the-letter-"s"

"keyup" events apparently don't have a useful `charCode` field so this
PR looks at the `keyCode` field instead

cc @iitalics


iitalics:
Sweet

PR #207, opened 2019-08-12
Create good-first-issue.md

spdegabrielle:
This issue template provides a means to easily record 'good first issue' issues with the 'good first issue' label, so they can be easily found by new and aspiring racketeers.

### DIFF
--- a/.github/ISSUE_TEMPLATE/good-first-issue.md
+++ b/.github/ISSUE_TEMPLATE/good-first-issue.md
@@ -1,0 +1,30 @@
+---
+name: Good first issue
+about: template for 'good first issue'
+title: ''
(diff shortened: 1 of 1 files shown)


sorawee:
"Good first issue", as it's used in most GitHub projects, is a label that the _project maintainers_ organizes. It's up to the project maintainers to judge which issue is easy to fix. Letting bug reporters to decide which issue is easy to fix might not be desirable. E.g., they might mislabel it, and this means project maintainers need spend time relabelling. In the worst case where most issues are labeled incorrectly, the label would lost its meaning, making it useless.

spdegabrielle:
agreed

PR #224, opened 2020-01-02
Fix the contracts on `exec`/`commandline` to match their behavior

lexi-lambda:
`exec` and `commandline` are documented as decoding their content, and they have contracts that suggest that they do, but in fact they do not. Whether originally intentional or not, this corrects the documentation and contracts to match their true behavior.

### DIFF
--- a/scribble-doc/scribblings/scribble/manual.scrbl
+++ b/scribble-doc/scribblings/scribble/manual.scrbl
@@ -1714,9 +1714,8 @@ Typesets the given combination of a GUI's menu and item name.}
 @tech{decode}d @racket[pre-content] as a file name (e.g., in
 typewriter font and in quotes).}
 
-@defproc[(exec [pre-content pre-content?] ...) element?]{Typesets the
-@tech{decode}d @racket[pre-content] as a command line (e.g., in
-typewriter font).}
+@defproc[(exec [content content?] ...) element?]{Typesets the
+@racket[content] as a command line (e.g., in typewriter font).}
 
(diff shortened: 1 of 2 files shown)


rfindler:
It looks like a commit you pushed recently, @lexi-lambda changed the behavior of those functions? The way the comments above are worded, however, don't seem to match with this? Can you clarify?

https://github.com/racket/racket/commit/07f060133cf4eb4df9968faa2b5521710fb271b4

lexi-lambda:
Sorry, I should have made the connection more clear.

The changes in https://github.com/racket/racket/commit/07f060133cf4eb4df9968faa2b5521710fb271b4 updated variants of `exec` and `commandline` that are [defined within the `racket-build-guide` package](https://github.com/racket/racket/blob/07f060133cf4eb4df9968faa2b5521710fb271b4/pkgs/racket-build-guide/common.rkt) in the main repository. However, those custom variants were inconsistent with the real versions (defined in `scribble-lib`), causing various command line flags to appear as `–flag` rather than `--flag`. https://github.com/racket/racket/commit/07f060133cf4eb4df9968faa2b5521710fb271b4 fixed the behavior of the variants to match the real ones, fixing the typesetting problem.

However, while doing that, I noticed that the real `exec` and `commandline` functions were *documented* as decoding their arguments, possibly causing the confusion in the first place. So I opened this PR to fix the documentation bug.

rfindler:
Thanks!.

rfindler:
Thanks!

Issue #228, opened 2020-04-29
Semantic overhaul of Scribble with RDFa


gregid:
This is a proposal for an overhaul of current Scribble implementation. 

_The goal of the project is for every document generated by Scribble to be semantically annotated with RDFa + JSON-LD and search to be (also) implemented with HDT/SPARQL as engine._

#### The benefits:
- Flexible Datalog like querying for every facet of document
- Possibility of building Natural Language Querying on top of RDF data and SPARQL- Ease of extending properties thanks to Open World Assumption of OWL.
- Data annotation can be as granular as needed (e.g. it could include source code line on github for every paragraph).
- Semantic annotation is accessible for search engines crawlers.
- If packagsrc would implement semantic data the data integration between these 2 would be seamless (e.g. you could search for all packages that use SomeFunction)

#### The means/Required work:
1. implement RDF toolset in Racket:
  - RDF turtle/NQuads
  - JSON-LD (implemented by @cwebber)
  - RDFa parser
  - HDT ffi (for data storage and querying)
2. Building OWL ontology for Scribble documentation/code
3. Rework of Scribble output to replace current class based annotation with RDFa. If you are not familiar with Semantic annotation take a look at https://schema.org/Article especially scroll to the bottom to the RDFa tab.
4. Data querying based on data stored in HDT files for local documentation and possibly triple store (e.g. RDF4J) for the server side documentation. http://www.rdfhdt.org/what-is-hdt/

If you have any questions regarding benefits/implementation details feel free to ask here.

jackfirth:
I'm a little confused, are you proposing this as something _you'd_ like to do? Or the current maintainers of Scribble? The latter is unlikely to happen since this is a very large project.

gregid:
I am writing it down so it won't be lost in a case I won't be able to come back to this myself and to open this for discussion and critical review. 

Referenced issue and pull request were just the trigger as with semantified scribble these issues could be easily solved (and more).
 
I would love to do this myself as this is my area of expertise. But as mentioned on Racket Users at the moment I can't afford to spend time on free work, at least until I start having some income from my business in the making.

And yes I agree this is a large project, but it is also possible it could interest someone as a research project in the Uni Departments dealing with semantic data/ontologies/knowledge graphs.

In the meantime the least I can do is discuss the proposal if there is any interest at all.

PR #229, opened 2020-04-30
Automatically link to source code of document.

### LINKED ISSUE #208: add source repository to scribble manual

samth:
Automatically link to source code of document.

This works, provided that:
  - the package is on the catalog at pkgs.racket-lang.org
  - the package is hosted on GitHub
  - the package source does not specify a branch

Restriction 3 is easy to lift with more parsing.
Restriction 2 could be lifted for other known sites (such as gitlab)
 or packages hosted as directories.
Restriction 1 would be harder to lift.

This only links to the _top level_ file that defines the overall
document, not the file defining the particular section. Fixing that
would require the Scribble renderer to provide more detailed
information, although it's certainly a possible extension.

Currently, some single-collection packages get the wrong answer,
but that will change when racket/pkg-index#25 finishes updating on
the server.

Closes #223
Closes #208
Closes racket/racket#874    
Related to #76, #228

### DIFF
--- a/scribble-lib/scribble/manual-racket.js
+++ b/scribble-lib/scribble/manual-racket.js
@@ -15,14 +15,170 @@ AddOnLoad(function() {
     }
 })
 
+// cache of source urls
+var cache = {};
+
+function ParseSource(source, mod_path, single_collection) {
+
+    var source_url = new URL(source);
+
+    if (source_url.protocol == "github:") {
+        // browser URL parser only works with http(s) URLs
+        source_url = new URL("https" + source.substring(6));
+        var host = source_url.host;
+        var url_path = source_url.pathname.substring(1).split("/");
+        if (!(url_path.length >= 2)) return null;
+        var user = url_path.shift();
+        var repo = url_path.shift();
+        var branch = url_path.shift();
+        var source_path = url_path.join("/");
+    }
+    else if (("https:" == source_url.protocol) || ("git:" == source_url.protocol)) {
+        // browser URL parser only works with http(s) URLs
+        if ("git:" == source_url.protocol)
+            source_url = new URL("https" + source.substring(3));
+
+        var host = source_url.host;
+        var source_path = source_url.searchParams.get("path");
+        var branch = (source_url.hash || "#master").substring(1);
+        var url_path = source_url.pathname.substring(1).split("/");
+        if (url_path.length < 2) throw [source_url.pathname, url_path];
+        var user = url_path.shift();
+        var repo = url_path.shift();
+        var mtch = repo.match(/(.*)\.git$/);
+        if (mtch) repo = mtch[1];
+
+    }
+    else return null;
+
+    var mod_path_re = /^\(lib "(.+)"\)$/;
+
+    var mod_path_elems = mod_path && mod_path.match(mod_path_re)[1].split("/");
+
+    if (!user || !repo || !mod_path_elems)
+        return null;
+    if (single_collection)
+        mod_path_elems.shift();
+
+    var file_path = mod_path_elems.join("/");
+
+
+    if (source_path) {
+        file_path = source_path + "/" + file_path;
+    }
+
+    return { user: user,
+             repo: repo,
+             file_path: file_path,
+             branch: branch,
+             host: host };
+}
+
+function AddSourceElement(pkg_url, info) {
+    info.appendChild(document.createTextNode("Document source "));
+    var url_line = document.createElement("div");
+    var a = document.createElement("a");
+    a.href = pkg_url;
+    a.style.whiteSpace = "nowrap";
+    a.appendChild(document.createTextNode(pkg_url));
+    addSpan(url_line, "\xA0", "RktRdr");
+    url_line.appendChild(a);
+    info.appendChild(url_line);
+}
+
+var prefixes = { "github.com": "tree",
+                 "gitlab.com": "-/blob" };
+
+
+function AddSourceUrl(source, mod_path, collection, info) {
+    // multi is encoded as an array, empty as false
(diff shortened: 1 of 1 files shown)


mflatt:
This worked for me for the Plait documentation, but not for most things.

For most things, if I look in the consolev, I see an error about `source_found` being null. Probably the script should not fail if the URL doesn't match. Meanwhile, the things that failed to match have "?path=", such as `git://github.com/racket/scribble/?path=scribble-doc`, which seems not to match due to the `/` before the `?`.

(Is a regular expression really the right way to parse a URL in Javascript?)

samth:
Can you try again now? I've fixed a few bugs recently.

mflatt:
Yes, this works for me.

scribble-lib/scribble/manual-racket.js
```diff
@@ -15,14 +15,130 @@ AddOnLoad(function() {
     }
 })
 
+// cache of source urls
+var cache = new Map();
+
+function ParseSource(source, mod_path, single_collection) {
+    let source_re = /^(https|git):\/\/\github.com\/([^\/]+)\/([^\?\/]+?)(\/|)(\.git|)(\?path=(.+)|)(#[A-Za-z0-9_-]*|)$/;
```
sorawee, line 22: `let` doesn't work on old browsers. If you care about backward compatibility, you might want to use `var` instead.

scribble-lib/scribble/manual-racket.js
```diff
@@ -15,14 +15,130 @@ AddOnLoad(function() {
...
+}
+
+
+function AddSourceUrl(source, mod_path, collection, info) {
+    // multi is encoded as an array, empty as false
+    single_collection = (typeof collection === "string");
+
+    let parsed = source && mod_path && ParseSource(source, mod_path, single_collection);
+
+    let correct_url = `https://github.com/${parsed.user}/${parsed.repo}/tree/${parsed.branch}/${parsed.file_path}`;
```
sorawee, line 71: Ditto for template literal.

samth:
@sorawee mentions old browser compatibility. I'm pretty sure both let and template literals are available for basically everyone who wants this feature, but maybe we should have a general policy about what we want to be compatible with.

mflatt:
That sounds fine for the new feature, as long as using potentially unsupported things doesn't cause a failure that breaks old features.

samth:
I think that both of these features would cause all of manual-racket.js to fail to parse.

scribble-lib/scribble/manual-racket.js
```diff
@@ -15,14 +15,130 @@ AddOnLoad(function() {
...
+    let file_path = mod_path_found && (single_collection ?
+                                       mod_path_found[2] :
+                                       mod_path_found[1] + "/" + mod_path_found[2]) ;
+
+
+    if (source_path) {
+        file_path = source_path + "/" + file_path;
+    }
+
+    return {user: source_user,
```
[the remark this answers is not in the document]
samth: I think this should be fixed now. Can you check?

samth:
On the browser-compatibility side, here's [let](https://caniuse.com/#feat=let), [template literals](https://caniuse.com/#feat=template-literals), and [fetch](https://caniuse.com/#feat=fetch). All are over 95% of desktop browsers, and I expect even higher for people who want to edit Scribble files.

Having `let` or templates not work would cause the whole file to fail to parse. Having `fetch` not available would cause the code to error before doing anything useful (including for the previous message for how to link to the document in Scribble). Probably `fetch` could be handled with an exception handler or a callback. 

Additionally, the code uses `Map`, which also doesn't work in pre-2015 browsers.

None of these changes would make rendering the actual document break on any browser. If people want, I can take all of these out, though.

samth:
And about regexps -- using the `URL` constructor would take care of separating `git://github.com/foo/bar.git?path=xxx#master` into 

`URL { href: "git://github.com.com/foo/bar/?path=xxx#master", origin: "null", protocol: "git:", username: "", password: "", host: "", hostname: "", port: "", pathname: "//github.com.com/foo/bar/", search: "?path=xxx" }`

but that's not that helpful.

sorawee:
Just FYI: I would be very happy if Racket/Scribble switches to modern JS. I just raised the point in case you really want to commit to support very old browsers.

samth:
I just rewrote using the browser URL parser, and handled old browsers (I think; it's hard to test) and added GitLab support. @bennn can you test again?

scribble-lib/scribble/manual-racket.js
```diff
@@ -15,14 +15,175 @@ AddOnLoad(function() {
...
+    // avoid failing on browser that don't have `fetch`
+    if (mod_path && source_pkg && tag && window.fetch) {
+
+        var cached = cache[mod_path]
+        if (cached) {
+            AddSourceElement(cached[0], mod_path, cached[1], info);
+        }
+        else {
+            fetch("https://pkgs.racket-lang.org/pkg/" + source_pkg + ".json")
+                .then(response => response.json())
```
sorawee, line 169: It looks like you want to commit to support old browsers after all. In that case, arrow function here will cause a parsing error in old browser.

bennn:
Looks good for basic GitLab urls (pict-abbrevs) & ones with a path (csexpr)

I couldn't find any pkgs with usable version exceptions on GitLab

spdegabrielle:
This is awesome - Thank you.

PR #232, opened 2020-05-20
Implement support for converting values to GIFs

### LINKED ISSUE #231: Support for converting values to GIFs
I'd like to write some documentation with example code that includes embedded animations. I assumed I'd be able to do this by making the animations implement the `prop:convertible` interface and support conversion to GIFs. But according to the docs for [`render-convertible-as`](https://docs.racket-lang.org/scribble/core.html?q=special#%28def._%28%28lib._scribble%2Fhtml-properties..rkt%29._render-convertible-as%29%29), it looks like only the PNG and SVG formats are supported. Is GIF support reasonable?

jackfirth:
Closes #231. This pull request only touches the HTML rendering implementation code; it has no tests and does not yet update the docs. I'd like to get some feedback on whether this is the right implementation approach before getting to tests and docs.

### DIFF
--- a/scribble-doc/scribblings/scribble/core.scrbl
+++ b/scribble-doc/scribblings/scribble/core.scrbl
@@ -1848,13 +1848,15 @@ For a @racket[part] that corresponds to an HTML page, adds content to
 the @tt{<head>} tag.}
 
 
-@defstruct[render-convertible-as ([types (listof (or/c 'png-bytes 'svg-bytes))])]{
+@defstruct[render-convertible-as ([types (listof (or/c 'png-bytes 'svg-bytes 'gif-bytes))])]{
  For a @racket[part] that corresponds to an HTML page,
  controls how objects that subscribe to the @racketmodname[file/convertible]
  protocol are rendered.
       
  The alternatives in the @racket[types] field are tried in order
  and the first one that succeeds is used in the html output.
+
+ @history[#:changed "1.34" @elem{Added support for  @racket['gif-bytes].}]
 }
 
 @defstruct[part-link-redirect ([url url?])]{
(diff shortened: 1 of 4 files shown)


jackfirth:
Manually tested this on my machine and it works. But I don't how to write an automated test for this, or which places in the documentation need to be updated. Suggestions greatly appreciated!

samth:
Looking at the existing tests, I don't see anything that tests any aspects of the html rendering. 

A number of places (core.scrbl, running.scrbl) seem to mention the supported formats. 

There's also the `--convert` flag, which should maybe also accept `gif`, and the `render-convertible-as` property. I don't know if it makes sense to change all these, but it might be worth looking at.

jackfirth:
I see a `++convert` flag, but not a `--convert` one. The `++convert` one already accepts gifs I think. From what I can decipher it's only the `prop:convertible` mechanism that doesn't accept gifs, images added with `@image{}` can be png, svg, or gif.

jackfirth:
Changed `render-convertible-as?` to allow gifs. Skipping tests since manual testing seems to work and I have no idea how to set up html rendering tests.

mflatt:
Ok by me. Is it worth extracting the GIF width and height to include as attributes in the `img`?

soegaard:
Den tor. 21. maj 2020 kl. 05.07 skrev Sam Tobin-Hochstadt <
notifications@github.com>:

> There's also the --convert flag, which should maybe also accept gif, and
> the render-convertible-as property. I don't know if it makes sense to
> change all these, but it might be worth looking at.
>
As far as I know, there is only one reason to use gifs namely to display
animations.
For all other purposes the existing image formats are better.
Therefore a general convert-to-gif option isn't worth it.

/Jens Axel

jackfirth:
@mflatt Yes, that's a good idea. Will update the PR.

jackfirth:
@mflatt Done!

PR #237, opened 2020-05-25
Allow full customization of `@other-doc` content.

samth:


mflatt:
I think of the generated content as being the point of `other-doc`, and `secref` lets you specify the content. But if `#:content` here is changed to always take effect (not just when `#:indirect?` is true), I guess this would be ok as a different way of spelling `(secref "top" ....)`.

PR #240, opened 2020-06-01
Add browser support policy for Scribble-generated documentation.

samth:
The short summary: all evergreen browsers from the past year,
plus anything <5 years old with at least 1% share on docs.racket-lang.org.

[Rendered](https://github.com/samth/scribble/blob/browser-support/scribble-lib/scribble/browser-support.md)

### DIFF
--- a/scribble-lib/scribble/browser-support.md
+++ b/scribble-lib/scribble/browser-support.md
@@ -1,0 +1,41 @@
+## Scribble Browser Support Policy
+
+This document describes the intended browser support range for HTML
+manuals generated with `#lang scribble/manual`.
+
+This policy defines three levels of support:
+
+- Full support (all features). 
+- Partial support (documentation features)
+- Minimal support (reading documentation text)
+
+Any feature must work correctly in all browsers with full support.
+
+Features that are not needed for reading and understanding the
+documentation, but that are intended for developers of the
+documentation or other specialized audiences, such as links to
+implementation, may not work in browsers with only partial support.
+
+For browsers with minimal support, it will be possible to read the
+text of the documentation, but all other features (such as proper
+styling or links) may be missing.
+
+#### Currently supported browsers
+
+- All "evergreen" browsers current within the past 12 months (this
+  includes Chrome, Firefox, Safari, Edge, and other regularly-updated
+  browsers based on similar). "Current" means if that browser was
+  shipped to users within the past 12 months, it is fully supported.
+  
+- Partial support includes any browser released within the past 5
+  years with at least 1% use on https://docs.racket-lang.org As of
+  June 1, 2020, that includes the above browsers only.
+  
+  Note that the only browser not in the above set with more than 1%
+  share is Chrome 38 as distributed on the Nexus 5, which was
+  discontinued in March 2015.
+  
+- Minimal support applies to all widely-distributed browsers. That is,
+  it is important that the text of the documentation appears in Links,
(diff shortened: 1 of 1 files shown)


scribble-lib/scribble/browser-support.md
```diff
@@ -0,0 +1,41 @@
...
+
+- Full support (all features). 
+- Partial support (documentation features)
+- Minimal support (reading documentation text)
+
+Any feature must work correctly in all browsers with full support.
+
+Features that are not needed for reading and understanding the
+documentation, but that are intended for developers of the
+documentation or other specialized audiences, such as links to
```
jackfirth, line 16: "such as links to implementation" - is this the motivating use case for this pull request? (Just curious)

sorawee:
I think the requirement to have a minimal support for IE6 doesn't make much sense. IE6 by default can't even view sites with https protocol. Even when TLS1.0 is manually enabled, racket-lang.org (and most websites in the world) seems to be serving a content that IE6 doesn't understand, so it will fail with "The page cannot be displayed". This means that users who are still using it can't download Racket in the first place, so there's really no point to support it.

sorawee:
FYI: if anyone wants to test IE (including IE6), you can use VMs provided by Microsoft. Unfortunately, Microsoft took these VMs down from their website for unknown reasons, but some users archived them. Use them at your own risk.

- https://gist.github.com/zmwangx/e728c56f428bc703c6f6#gistcomment-3196040
- https://gist.github.com/zmwangx/e728c56f428bc703c6f6#gistcomment-3115797

As noted above, for IE6, we need to enable TLS1.0 to access https websites, but racket-lang.org can't be accessed regardless.

LiberalArtist:
FWIW, while I don't care about IE6, I do use `lynx` from time to time, and the docs are quite readable in it.

scribble-lib/scribble/browser-support.md
```diff
@@ -0,0 +1,43 @@
...
+For browsers with minimal support, it will be possible to read the
+text of the documentation, but all other features (such as proper
+styling or links) may be missing.
+
+#### Currently supported browsers
+
+- All "evergreen" browsers current within the past 12 months (this
+  includes Chrome, Firefox, Safari, Edge, and other regularly-updated
+  browsers based on similar). "Current" means if that browser was
+  shipped to users within the past 12 months, it is fully supported.
```
LiberalArtist, line 28: Is the intention that this should include supported Firefox ESR releases, since they get point releases (that are not feature updates) more than once per year? https://support.mozilla.org/en-US/kb/firefox-esr-release-cycle

scribble-lib/scribble/browser-support.md
```diff
@@ -0,0 +1,43 @@
...
+  evergreen browsers; the current usage share should be consulted when
+  deciding whether a specific older browser still qualifies.
+
+- Minimal support applies to all widely-distributed browsers that can
+  reach the documentation. That is, it is important that the text of
+  the documentation is readable in text-mode browsers such as `lynx`
+  and `links`, in screen readers, and in other browsers with limited
+  styling or scripting. (Very old browsers that can no longer establish
+  a modern TLS connection to the documentation site, such as IE 6, are
+  out of scope: they cannot load the pages at all.)
```
LiberalArtist, line 42: I am not against declaring IE 6 out of scope, but the TLS criterion is about accessing documentation over the network. Isn't IE 6 still able to load locally-installed documentation pages?

PR #242, opened 2020-06-07
[WIP] Support reference to page numbers of an element tag

chihyang:
This is a pull request **working in progress**.

For docs such as pdf, page number reference is sometimes very useful. To support page reference, a new kind of tag is added to scribble/base. Besides, a new command `PageRef` is added to scribble.tex. Its argument is a page number. By default, it shows 'page' before the page number. Users can redefine this command to get different outputs.


chihyang:
This is a pull request **working in progress**. And I am not sure if it is suitable for merging into the main scribble repo. So I put it here for discussion.

Support for page reference in PDF is very useful. Support for page reference in HTML or other formats seems less meaningful. Because most of time there is no fix-sized paper in such documents (maybe this is why `pageref` is not added to scribble code?). In such format, `pageref` should be rendered as a hyperlink with a configurable text, such as "here".

More generally, this is a problem about how to use tags conveniently: how can I use something like LaTeX `\ref` in scribble, and make it work in all the supported documents? Now `enconde-tag` is implemented in different renders, so this seems a little hard. But if there is a way to use the *real* content of a tag in the document directly, things like `pageref` can be avoided.

For example, once we can use the "real" content of a tag, I can use the following style:

``` tex
\newcommand{\PageRef}[1]{page~\pageref{#1}}
```

then in the style file:

``` racket
(define PageRef
  (make-style "PageRef" (list (make-tex-addition "pageref.tex"))))
```

then in the scribble file:

``` racket
;;; tag here
(elemtag "a-tag")

;;; use the tag somewhere else
(elem #:style PageRef (encode-tag "a-tag"))
```

Because `encode-tag` is extracts the content of the tag, it is also possible to
define different styles for different formats.

PR #246, opened 2020-06-19
Fix #131: disallow bogus mode arguments and fix related bugs

### LINKED ISSUE #131: defmethod accepts bogus mode arguments
The `defmethod` form silently ignores bad `#:mode` arguments (e.g. `extends` rather than `extend`). It would be nice for it to raise an error, instead.

sorawee:
- Raise a syntax error when bogus mode arguments are given.
  Note that https://github.com/racket/gui/pull/193 must be merged first.
- Fix a bug where `extend-final`, `augment-final`, and `public-final` are not recognized
- Document the `public` mode which was already supported but not
  documented.
- Clarify the default mode when `maybe-mode` is not given.

### DIFF
--- a/scribble-doc/scribblings/scribble/manual.scrbl
+++ b/scribble-doc/scribblings/scribble/manual.scrbl
@@ -1609,14 +1609,15 @@ accepted and propagated to the superclass.}
               ([maybe-link code:blank
                            (code:line #:link-target? link-target?-expr)]
                [maybe-mode code:blank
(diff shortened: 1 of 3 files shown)


sorawee:
This still isn't quite right because when `public-final` is given, `finality` is ignored.

sorawee:
Should work now.

mflatt:
Thanks!

PR #250, opened 2020-07-05
make the danger symbol (⚠ U+26A0) work in the latex backend

rfindler:
using one of the approaches found here
https://tex.stackexchange.com/questions/159669/how-to-print-a-warning-sign-triangle-with-exclamation-point

This fixes https://github.com/Metaxal/quickscript/issues/31 but I'm not sure if it is the best way.

### DIFF
--- a/scribble-lib/scribble/latex-render.rkt
+++ b/scribble-lib/scribble/latex-render.rkt
@@ -1321,6 +1321,7 @@
                                 [(#\⟜) "$\\multimapinv$"]
                                 [(#\⅋) "$\\invamp$"]
                                 [(#\□) "$\\square$"]
+                                [(#\⚠) "{\\fontencoding{U}\\fontfamily{futs}\\selectfont\\char 66\\relax}"]
                                 [else
                                  (cond
                                   [(char<=? #\uAC00 c #\uD7AF) ; Korean Hangul


Metaxal:
Another possibility is to display a small png maybe?

rfindler:
I think that constructing a pict could work; doing the equivalent of this latex code: `\stackon[1.3pt]{\color{red}$\triangle$}{\tiny !}}` (as suggested at that same link).

rfindler:
.... although I do see that there is currently no dependency between the scribble and pict packages. So maybe setting up a danger thingy that implements the `file/convertible` protocol for the types listed here: https://github.com/racket/scribble/blob/master/scribble-lib/scribble/latex-render.rkt#L66-L74

rfindler:
I guess that that approach is probably not worth it unless there's something that's explicitly wrong with the latex code, however.

Metaxal:
Yeah, otherwise don't bother, I'll just remove it entirely, and use `@image` to display a small png, or just nothing at all.

rfindler:
It might be easier (and it would probably look better) to use a pict instead of an image.

Metaxal:
But a pict would give me a dependency on `racket/pict`.
I'm thinking of saving a large enough picture such as:
```racket
(send (pict->bitmap (text "⚠" '() 128))
        save-file
        "img/warning.png"
        'png)
```
and reducing it in the text.

rfindler:
Sure. But a dependency on pict seems fine for the manual. Many manuals have that (as does drracket). 

Also, the image won't scale nicely in html or in pdf, but the pict will.

The code turned out to be more complex than I thought! Anyway, feel free to use or not.

```
#lang racket
(require pict)
(define (triangle p)
  (define w (pict-width p))
  (define h (pict-height p))
  (dc
   (λ (dc dx dy)
     (define left 0)
     (define right w)
     (define bot (* h 1.1))
     (define top (* h .2))
     (define (dr x1 y1 x2 y2)
       (send dc draw-line
             (+ dx x1) (+ dy y1)
             (+ dx x2) (+ dy y2)))
     (dr left bot (* 1/2 (- right left)) top)
     (dr (* 1/2 (- right left)) top right bot)
     (dr left bot right bot))
   w h))
(define bang (text "!"))
(define p (triangle (text "w")))
(define warning
  (refocus
   (cbl-superimpose p
                    (scale bang .7))
   p))

;; see it in the drracket repl (in dark mode; drop the colorize if not in dark mode)
(colorize
 (scale
  (hbl-append
   (text "abc")
   warning
   (text "def"))
  10)
 "white")
```

Metaxal:
Thanks! But to use a pict in scribble, don't you have to turn it into an image?

rfindler:
No, you just drop it there like you can drop strings. Scribble does the rest.

And actually, this is probably going to be slightly better (the turns in the triangle will look nicer with this code I believe).

```
(define (triangle p)
  (define w (pict-width p))
  (define h (pict-height p))
  (define left 0)
  (define right w)
  (define bot (* h 1.1))
  (define top (* h .2))
  (define points
    (list (cons left bot)
          (cons (* 1/2 (- right left)) top)
          (cons right bot)))
  (dc
   (λ (dc dx dy)
     (define brush (send dc get-brush))
     (send dc set-brush "black" 'transparent)
     (send dc draw-polygon points dx dy)
     (send dc set-brush brush))
   w h))
(define bang (text "!"))
(define p (triangle (text "w")))
(define warning
  (refocus
   (cbl-superimpose p
                    (scale bang .7))
   p))
```

rfindler:
Maybe we can leave well enough alone here.

PR #252, opened 2020-07-15
Improve tech normalization of plural words

jackfirth:
This pull request adjusts the behavior of `@tech{}` to better handle normalizing plural words to their singular forms. Specifically, it adds a rule to strip a trailing "es" in some circumstances and to **not** strip a trailing "s" if the last two letters of the term are "ss". I wanted this because `@tech{syntax class}` and `@tech{syntax classes}` currently normalize to `syntax clas` and `syntax classe` respectively.

### DIFF
--- a/scribble-doc/scribblings/scribble/manual.scrbl
+++ b/scribble-doc/scribblings/scribble/manual.scrbl
@@ -1791,7 +1791,10 @@ normalized as follows:
 
  @item{A trailing ``ies'' is replaced by ``y''.}
 
- @item{A trailing ``s'' is removed.}
+ @item{A trailing ``es'' is removed if it was preceded by an ``s'', ``x'',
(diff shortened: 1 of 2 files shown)


mflatt:
My only hesitation here is that this change can break some cross-document references. For example, someone might have rendered a paper that points to documentation at `https://docs.racket-lang.org/` on the theory that tag links are stable. I don't know how big of a problem that might be, though. Any thoughts?

I've sometimes had to write things like `@tech{syntax class}es` to get links to work, so I can certainly appreciate the improvement.

jackfirth:
It's possible. I don't know of a good fix for that problem. But I don't think "we can never change anything about how tag links are created" is an acceptable stalemate. Is there a way we can check that this at least doesn't break any documentation links in existing packages?

samth:
We could just generate both tags in the HTML.

PR #257, opened 2020-08-02
Fix #255: update Fira Mono

### LINKED ISSUE #255: Update Fira Mono
"Fira Mono" that Scribble is currently using is not truly monospace. For an example, see the unicode table at https://docs.racket-lang.org/text-table/ which displays incorrectly.

I tried to update "Fira Mono" to its most recent version at https://fonts.google.com/specimen/Fira+Mono (Note: I haven't tried the ones in https://github.com/mozilla/Fira, but I believe the repo is outdated and no longer updated), and found that it does fix the issue: the font is now truly monospace.

I'm not familiar with the base64 encoding process and not sure how the update would affect the license, so I'd probably leave this to @mbutterick.

sorawee:
Before (https://docs.racket-lang.org/text-table/index.html):

Screen Shot 2020-08-02 at 02 18 29

After:

Screen Shot 2020-08-02 at 02 18 35

Before (https://docs.racket-lang.org/2d/):

Screen Shot 2020-08-02 at 02 16 24

After:

Screen Shot 2020-08-02 at 02 17 17


rfindler:
Yay!

PR #258, opened 2020-08-06
Add racketoptionalfont

sorawee:
Scribble used to support a custom style for optional brackets.
In particular, the optional brackets will be given the `'paren-shape`
of value `#\?`[1], and the CSS class `opt-color` will be tagged
to these brackets.

Later, Scribble supports the curry notation. Its implementation
no longer uses `'paren-shape` to indicate optional brackets.
Unfortunately, it also accidentally dropped the `opt-color` tagging[2].

This PR restores the original behavior by tagging the class
`opt-color` to optional brackets. It also adds `racketoptionalfont`
so that Scribble users can typeset optional brackets.
Lastly, it cleans up the code that supports the `'paren-shape` of value
`#\?`, since it is effectively a deadcode.

Note that this PR does _not_ change any CSS styling, so there's no
visible change. It would make CSS styling customization easier, however.

[1]: https://github.com/racket/scribble/commit/9b7993ea02d8f90f176e6305e8e1e7e46f5d4384#diff-017add06555fc85fa3ae5f27a3eb52cbR253
[2]: https://github.com/racket/scribble/commit/95ecb101d1cc61d212c4d520#diff-017add06555fc85fa3ae5f27a3eb52cbR879

### DIFF
--- a/scribble-doc/scribblings/scribble/demo-manual.scrbl
+++ b/scribble-doc/scribblings/scribble/demo-manual.scrbl
@@ -8,10 +8,6 @@
 
 @(define css tt)
 
-@(define-syntax (opt-example stx)
-  ;; A #\? 'paren-shape value triggers RktOpt:
-   #`@racket[#,(syntax-property #'(in-example) 'paren-shape '#\?)])
-
 @title{Manual All-Styles Document}
 
 @table-of-contents[]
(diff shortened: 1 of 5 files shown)


scribble-doc/scribblings/scribble/demo-manual.scrbl
```diff
@@ -76,7 +72,9 @@
 
  @item{@css{RktModLink} (a linked module reference): @racketmodname[racket/base]}
 
- @item{@css{RktOpt} (option-argument brackets): brackets in @opt-example[]}
+ @item{@css{RktOpt} (option-argument brackets): brackets in
+
+       @defproc[(f [x any/c 1]) any]}
```
mflatt, line 77: This will create a "no declared exporting libraries for definition". I think there must be some way to render a `defproc` without link trouble, but I'm not sure of the right combination. Probably `#:link-target? #f` is part of the answer.

Issue #261, opened 2020-08-18
autobib renders hash rather than citation


wilbowma:
`meow.scrbl`
```
#lang scribble/base
@; scribble --pdf meow.scrbl

@(require
  scriblib/bibtex)

@(define-bibtex-cite "wilbowma.bib" ~cite citet generate-bibliography)

@title{Test}

@citet{matthews2007} is cool stuff.

@(generate-bibliography)
```
`wilbowma.bib`
```
@Inproceedings{matthews2007,
  author    = {Matthews, Jacob And Findler, Robert Bruce},
  title     = {Operational Semantics for Multi-language Programs},
  booktitle = popl,
  year      = {2007},
  doi       = {10.1145/1190216.1190220},
  file      = {:matthews2007 - Operational Semantics for Multi-language Programs.pdf:PDF},
}
```

![image](https://user-images.githubusercontent.com/1133598/90489615-57533600-e0f2-11ea-9dce-a24f4ea2a847.png)

wilbowma:
Looks like the problem is that the type is case sensitive.

wilbowma:
I think this should have been closed at the same time as https://github.com/racket/scribble/pull/264

PR #264, opened 2020-08-18
Fix case sensitivity issues in bibtex

wilbowma:
bibtex is supposed to treat keys and labels as case insensitive, according to what I can find (http://maverick.inria.fr/~Xavier.Decoret/resources/xdkbibtex/bibtex_summary.html#stringdef). This was mostly true already, but labels and entry keys were not treated this way. This patch fixes it.

### DIFF
--- a/scribble-lib/scriblib/bibtex.rkt
+++ b/scribble-lib/scriblib/bibtex.rkt
@@ -83,7 +83,7 @@
       [typ
        (read-char ip)
        (slurp-whitespace ip)
-       (define label (read-until (λ (c) (char=? c #\,)) ip))
+       (define label (string-downcase (read-until (λ (c) (char=? c #\,)) ip)))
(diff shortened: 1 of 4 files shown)


wilbowma:
Closes #261

mflatt:
I'm happy to merge this, but checking on one detail first: Using `string-foldcase` is normally more correct than `string-downcase` for setting up a case-insensitive comparison. It doesn't matter for English/Latin/ASCII letters, and really you want to do the same as bibtex, which may well be downcase. Do you want to keep downcase or change to foldcase?

wilbowma:
Should probably use `string-foldcase`. I was following the existing patterns in the file, which uses `string-downcase`; I didn't realize I ought to use `string-foldcase`. However, this means the other uses should probably be changed.

Issue #260, opened 2020-08-13
'top cell-property does not work for cells with picts in PDF backend


wilbowma:
Here's an example:
```
#lang scribble/base

@tabular[
#:row-properties '((top-border bottom-border))
(list
(list "> " @verbatim|{a very tall



cell.}|))
]

@tabular[
#:row-properties '((top-border top bottom-border))
(list
  (list "> " @verbatim|{a very tall



cell.}|))
]


@(require pict)
@tabular[
#:row-properties '((top-border top bottom-border))
(list
  (list "> " (filled-ellipse 40 40)))
]
```

Expected output, and the output in HTML:
![Screenshot_20200812_183954](https://user-images.githubusercontent.com/1133598/90084914-7e150500-dccb-11ea-822a-292f741d09e9.png)

Actual output in PDF:
![Screenshot_20200812_184043](https://user-images.githubusercontent.com/1133598/90084917-80775f00-dccb-11ea-966f-fcca46b0023f.png)

wilbowma:
Looks like the issue is in the way picts are rendered, as `includegraphic` doesn't have its box adjusted. The solutions I've found by playing with the latex and based on this article https://tex.stackexchange.com/questions/30079/how-to-align-picture-top-left-in-a-table

- use `\usepackage[export]{adjustbox}`, and and somehow wrap the cell in `\adjustbox{valign=t}{...}`. This might be undesirable if we don't want another dependency, and not sure if this can be done for all cells or if we need to detect a pict cell. This works very well in my tests and I get the same output as in the HTML. It also seems to be safe to wrap elements that have already been adjusted without doing any harm.
- change the `\raisebox` in render to use `\raisebox{-\height}`, if we're supposed to be top-aligned. Seems that `render-content` doesn't have that information, so not exactly sure how to implement that change. This also doesn't render exactly like the HTML; there's some extra space at the top of the cell (similar to the extra space at the bottom in the PDF example above)

wilbowma:
Using `adjustbox` doesn't quite get is to the same behavior on both backends. In this example, all look the same in HTML, but the first two are slightly different than the second two in PDF. Not sure who is to blame

```
#lang scribble/base
@tabular[
#:row-properties '((top-border top bottom-border))
(list
(list "> " @verbatim|{a very tall



cell.}|))
]


@(require pict)
@tabular[
#:row-properties '((top-border top bottom-border))
(list
(list "> " (filled-ellipse 40 40)))
]

@tabular[
[...]

PR #265, opened 2020-08-19
Attempt to fix verticle alignment of some boxes in latex backend

wilbowma:
This addresses #260 by providing a default box-adjuster. This requires an extra LaTeX package.

I'm not really sure this is a good solution, but I thought I'd offer a patch for experts to review. It might have unintended consequences if existing code was relying on ignoring the vertical alignment style property. The tests seem to pass, but it seems like a hard thing to write a test for.

### DIFF
--- a/scribble-lib/scribble/latex-render.rkt
+++ b/scribble-lib/scribble/latex-render.rkt
@@ -282,7 +282,7 @@
                                      ((box-mode-selector as-box-mode) a)))
                               (style-properties
                                (paragraph-style p)))
-                       "hbox")]
+                       (format "adjustbox{valign=~a}" as-box-mode))]
                      [(eq? sn 'author) "SAuthor"]
                      [(eq? sn 'pretitle) #f]
                      [(eq? sn 'wraps) #f]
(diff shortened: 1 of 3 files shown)


mflatt:
Adding a new package is always worrying, but I'd be happy to have a solution here.

Would using `minimal` instead of `export` work and potentially avoid unwanted interactions with uses of `includegraphics`, or is `export` reasonably unlikely to break existing documents?

@rfindler Do you think we should give this a try and check the document rendering and a few papers?

rfindler:
I tried the documentation PDF build and it caused no problems there. I also built a recent paper and it didn't cause any problems there either. I'm not savvy enough to understand how this fixes the problem but it sounds like you both are so I'm in favor!

wilbowma:
`minimal` is probably better; I was explicitly trying to avoid using it directly with `includegraphics` directly.

@rfindler, what was happening was the call to `do-render-paragraph` did specify the box should be top aligned by passing `'t` as `as-box-mode`. However, for paragraphs, this gets implemented by reading the style properties and looking for a `box-mode` implementation. The `pict` didn't have one, so it was wrapped in `hbox`, the default, which coverts to a box but doesn't adjust vertical alignment. This patch uses `adjustbox` as the default, which both converts to a box and adjusts vertical alignment.

wilbowma:
When I try `minimal`, I get complaints that `valign` is undefined by `adjustbox`, which is strange.

I also notice a small difference, although I haven't observed it yet: `\adjustbox{valign=c}` should probably behave like `\hbox`, since that's what it's replacing. But I think it behaves like `\mbox`. Unfortunately, I don't really understand how these two things differ.

wilbowma:
The vertical alignment in tables does not behave as expected. It seems to be only coincidence that my example worked. I'll continue debugging and post in #260, but this patch doesn't really work.

mflatt:
Thanks for investigating more. My vague memory of table-cell alignment was that the CSS and Latex models were deeply different, and I found some approximation that worked for some common cases and gave up on the rest.

wilbowma:
Yeah thats what im discovering totally understand why you gave up.

--

PR #268, opened 2020-08-19
Added support for latex escapes in bibtex titles and authors

wilbowma:
This allows using certain escape sequences, such as \", in titles and authors of
bibtex files.
Titles are wrapped titles with 'exact-chars, so the set of escapes handled is
large.

A different method is used for author which basically convert certain well-known
escapes into Unicode.
This seems necessary to support parsing of author names.


wilbowma:
FYI, if you end up wanting to merge all of these, I've merged them all in my own dev branch https://github.com/wilbowma/scribble/tree/dev

mflatt:
@jeapostrophe I think you originally wrote "bibtex.rkt". Any opinions before we merge? See also #267 and #264.

PR #270, opened 2020-08-31
add current-html-render-head-prefix

jeapostrophe:


jeapostrophe:
I'm working on a Scribble site where I want to put Google Analytics on every page.

I know I can do this with a `head-extra`, but it is difficult to make sure I include it on every single part. I was hoping that I could use some global configuration option to do this, but looking at

https://docs.racket-lang.org/scribble/config-style.html

I can only add files to the output, add CSS, and add to the prefix. Since the prefix is before the HTML tag, I can't use it to insert things into the head.
I tried to hack the render-mixin to automatically insert on render-part, but changing the style on the part made references not work. I also think this wouldn't be a good way for non-Racket hackers to do it. Similarly, I'm pretty sure it would make the script appear multiple times if a page had multiple parts.
Instead, I did this. If you like it, I'll add docs.

jeapostrophe:
Ping @mflatt

mflatt:
Sorry that I missed this, even with the ping.

I like the idea of a way to set a head addition everywhere, but I think it still works better to make it a property — one that gets propagated automatically to nested parts that are on their own page, or even lifted to the enclosing page if the property is no a same-page nested section. That's how `css-addition` and similar work, so you can use the existing `exrtact` traversal helper. The style-property approach tends to work better for Scribble documents, because there's generally not a good place to put a `parameterize` (and mutating a parameter has its own issues).

mflatt:
... and maybe `head-extra` should have worked that way in the first place.

Rechecking the documentation, I don't think it's clear on `css-addition`. It talks about lifting CSS additions up on the same page, but not about propagating to other pages for nested sections. I'll improve the docs.

mflatt:
Since I ended up spending time in this area, here's a specific proposal: https://github.com/mflatt/scribble/commit/7c7e1213d039bebde7e553b9a74a8329216d2799

jeapostrophe:
Thank you! I think this is good and would solve my problems. <3

If you're touching this code, something related that I'd like Scribble
to have built-in is a way to deal with this problem:

https://github.com/wilbowma/scribble-minted/commit/12360be85892a0204b89e98445fdb9ef6e41eb28

With scribble-minted, every minted block has to include a css file for
the particular language stylesheet, and this ends up with hundreds of
`style` blocks.

--
Jay McCarthy
Associate Professor @ CS @ UMass Lowell
http://jeapostrophe.github.io
Vincit qui se vincit.

PR #272, opened 2020-09-08
Disable txfonts package in scribble/acmart

### LINKED ISSUE #271: Adding txfonts Changes the Font of scribble/acmart
The inclusion of `txfonts` in c62d008cfb changes the font of acmart acmsmall articles. Could it be disabled for acmart files?

I am not sure if adding

```
\renewcommand\packageTxfonts\relax
```

to `scribble-lib/scribble/acmart/acmart-load.tex` would be a bad idea or not.

shhyou:
Commit c62d008cf include txfonts into the set of
default LaTeX packages. However, txfonts changes
the main font of the document, conflicting the
requirement of `scribble/acmart` articles.

We skip txfonts by setting `packageTxfonts` to nop.

Closes #271.

### DIFF
--- a/scribble-lib/scribble/acmart/acmart-load.tex
+++ b/scribble-lib/scribble/acmart/acmart-load.tex
@@ -5,6 +5,7 @@
 \renewcommand\packageTocstyle\relax
 \renewcommand\packageMathabx{\ifx\bigtimes\undefined \usepackage{mathabx} \else \relax \fi}
 % Both 'mathabx' and 'newtxmath' (required by the 'acmart' class) define a '\bigtimes' command. 
+\renewcommand\packageTxfonts\relax
 \let\Footnote\undefined
 \let\captionwidth\undefined
 \renewcommand{\renewrmdefault}{}


rfindler:
Looking at the docs for txfonts I don't see a way to load those new symbols but avoid changing the default fonts. Without that, some unicode characters will cause scribble to generate buggy latex (in scribble/acmart).

shhyou:
Replacing txfonts with newtxmath for those 4 symbols could be another workaround though I don't know if it would introduce other glitches.

shhyou:
I found that `txfonts` is superceded by `newtx` package (from, e.g. [here](https://tex.stackexchange.com/questions/81772/packages-times-itsc-txfonts-newtxtext-newtxmath-and-mathptmx-relation-and) or the official documentation). In `newtx`, the math symbols are provided by `newtxmath` and `acmart.cls` already included it. Therefore this change is good to go.

A separate issue is that scribble should use `newtx` instead of `txfonts`, but that is independent of this pull request.

shhyou:
An example to confirm this:

- `art.scrbl`

```
#lang scribble/acmart

@title{This is a Title}
@author{A1}

@abstract{
  abstraction matters
}

@section{Section 1}

test ⊸ ⟜ ⅋  □
```

- Steps:

```
scribble --latex art.scrbl
sed -e 's/\\usepackage{txfonts}/\\relax/' art.tex > art-notxfonts.tex
pdflatex art-notxfonts.tex
```

rfindler:
Thanks!

And I think it would be great to follow up on your suggestion of changing scribble too.

PR #274, opened 2020-09-19
Replace txfonts by txmath

shhyou:
Replace txfonts with newtxmath.

- The newtx package supersedes txfonts.

- As a side effect, we have to load amsmath
  as well and it has to be loaded before
  mathabx and wasysym.

- Add nointegrals option to wasysym to avoid conflicting amsmath.

- Pick the \widering from newtxmath over the \widering from mathabx.

For now, I have tried rendering PDFs for all languages but `scribble/jfp` in [High-Level Scribble API](https://docs.racket-lang.org/scribble/generic-prose.html) on the scribble files here: [scribble-langs-client.zip](https://github.com/racket/scribble/files/5248862/scribble-langs-client.zip).

The newtx package provides quite some [options](http://texdoc.net/texmf-dist/doc/fonts/newtx/newtxdoc.pdf) for selecting different styles of math symbols (e.g. `libertine`, `upint` or `smallerops`). I don't know which ones are more common. The decision is left to the user.

### DIFF
--- a/scribble-doc/scribblings/scribble/config.scrbl
+++ b/scribble-doc/scribblings/scribble/config.scrbl
@@ -524,9 +524,10 @@ style:
             @item{@ltxd[0]{packageHyperref}, defaults to @tt{\usepackage{hyperref}}}
             @item{@ltxd[0]{renewrmdefault}, defaults to @tt{\renewcommand{\rmdefault}{ptm}}}
             @item{@ltxd[0]{packageRelsize}, defaults to @tt{\usepackage{relsize}}}
+            @item{@ltxd[0]{packageAmsmath}, defaults to @tt{\usepackage{amsmath}}}
             @item{@ltxd[0]{packageMathabx}, defaults to @tt{\usepackage{mathabx}}}
-            @item{@ltxd[0]{packageWasysym}, defaults to @tt{\let\leftmoon\relax \let\rightmoon\relax \let\fullmoon\relax \let\newmoon\relax \let\diameter\relax \usepackage{wasysym}}}
-            @item{@ltxd[0]{packageTxfonts}, defaults to @tt{\usepackage{txfonts}}}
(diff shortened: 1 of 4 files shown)


shhyou:
(Caveat: adding latex packages to the default list is generally not backwards compatible.)

rfindler:
The docs all build in PDF form too. Is there anything else should try before merging? (Also I think it is our convention to put the docs and version number bump into the same commit as the main change.)

shhyou:
That's all I can think of, too. Nevertheless, it is to be expected that some existing documents will look different and some scribble files out there that have more customization can break.

We should include the commits for this issue in the release announcement. Is there any other way to make the changes more visible to the users?

rfindler:
`@history[]` annotations is all I can think of. I see that you have one already but it isn't really about the possible changes to the way documents render.

shhyou:
Okay, I added some pointers and link to this pull request.

shhyou:
Here are some examples documents rendered using scribble 1.29, 1.36 and 1.37.
[pdfs.zip](https://github.com/racket/scribble/files/5253129/pdfs.zip)

Source:

```
#lang scribble/base

@(require (only-in scribble/core element style))

@title{Title}
@author{Author}
@section{Header 1}

some content

@(element (style #f '(exact-chars))
          "$\\alpha^2+a^2+b^2+c^2=d^2 \\quad \\int \\iint$")

@(element (style #f '(exact-chars))
          "\\[\\alpha^2+a^2+b^2+c^2=d^2 \\quad \\int \\iint\\]")

@(element (style #f '(exact-chars))
          "$\\widering{IIIIIIIIIIIIIIIIIII}$")
```

rfindler:
Well, it looks to me like base-137 is a big improvement over base-136 and base-129 and art-137 is also an improvement over art-136 and art-129! Makes me feel pretty good about merging at this point so I'll just do that now.

PR #275, opened 2020-09-19
Block scribble from changing the font to ptm in LNCS

shhyou:
Block scribble from changing the default font to ptm for LNCS articles. Unfortunately, the math fonts are still incorrectly altered by newtxmath and I don't see a way to work around it.

Depends on #274.

### DIFF
--- a/scribble-lib/scribble/lncs/lang.rkt
+++ b/scribble-lib/scribble/lncs/lang.rkt
@@ -50,7 +50,8 @@
                                                     unicode-encoding-packages))
                 (scribble-file "lncs/style.tex")
                 (list cls-file)
-                #f))
+                #f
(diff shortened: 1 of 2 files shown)


rfindler:
LGTM

rfindler:
What's the state over here?  Would it work to add definitions of \multimap, \multimapinv, \invamp and \square into the latex file? I don't know latex well enough to know how to do it, but maybe there is a way to pull those definitions in without setting the font for everything? Or maybe they could be defined to be something that gives an error message telling people to go look at the file `lncs-load.tex` for a fuller explanation?

shhyou:
The main font is now Computer Modern, while the math font remains to be Times Roman. THe four new symbols can be rendered.

I don't know how to pull only those 4 symbols in. It is possible to disable the new math font entirely and define commands to emit error messages. I didn't touch newtxmath in this pull request.

PR #277, opened 2020-09-29
Fix JS error when searchbox is not present

sorawee:
When the search box is not present (e.g.,
https://docs.racket-lang.org/demo-m1/index.html),
pressing "S" will result in a JS error. This PR fixes the problem.
Note that semantically it makes more sense to give an ID to the
search box as we know exactly what search box we want.


mflatt:
@sorawee Looks like I created a conflict by merging another PR. It's probably very easy to fix, but can you do that, so I can squash-merge normally?

sorawee:
Yes, I will do that by the end of today.

sorawee:
@mflatt conflict resolved. Sorry for the delay.

Issue #206, opened 2019-08-11
Change scribble/blueboxes to scribble/signature-box


jackfirth:
In racket/drracket#291 and racket/drracket#172, it was decided that "signature box" is a much better term than "blue box", and various parts of drracket's documentation were updated to use this phrase. Scribble's [`scribble/blueboxes`](https://docs.racket-lang.org/scribble/blueboxes.html) library uses the term "blue box" in its documentation and the names of its exported identifiers.

We should add a `scribble/signature-box` library that's the same as `scribble/blueboxes` but with changed names. Then we should deprecate the `scribble/blueboxes` library and point to the new library in its documentation.

spdegabrielle:
@jackfirth I'd suggest this issue should be labelled 'good first issue' as it is a lot of steps but achievable for a beginner.

Steps to complete the task;
1. fork this repo`racket/scribble`
2. identify files that need to be changed <https://github.com/racket/scribble/search?q=blueboxes&unscoped_q=blueboxes>
3. create local copy on your PC
4. copy scribble-lib/scribble/blueboxes.rkt to scribble-lib/scribble/signature-box.rkt
5. copy `scribble-doc/scribblings/scribble/blueboxes.scrbl` to `scribble-doc/scribblings/scribble/signature-box.scrbl`
6. copy `scribble-lib/scribble/valid-blueboxes-info.rkt` to `scribble-lib/scribble/valid-signature-box-info.rkt`
7. change 'bluboxes' to 'signature-box' in files 4 ,5 , 6(above), `scribble-doc/scribblings/scribble/internals.scrbl`, `scribble-lib/scribble/contract-render.rkt` & `scribble-doc/scribblings/scribble/renderer.scrbl`.
8. delete `scribble-lib/scribble/blueboxes.rkt`, `scribble-doc/scribblings/scribble/blueboxes.scrbl` & `scribble-lib/scribble/valid-blueboxes-info.rkt`
9. test (is rebuilding racket/scribble sufficient?)
10. create PR for `racket/scribble`

jackfirth:
@spdegabrielle Good idea! I don't seem to have issue-labeling powers though, and I don't know who would.

jackfirth:
Thank you @rfindler!

nloadholtes:
Hi, I gave this a PR a try but the tests weren't passing. This is the first time I've tried working with racket so I'm not sure what the error message means. I put the details in the PR #286

PR #286, opened 2020-12-31
Refactor blueboxes name to signature-box

nloadholtes:
Simple find-and-replace to change the name for blueboxes to signature-box. 

(Quick note: this is my first time messing with racket, so apologies if anything is incorrect/wonky/etc.)

Followed the steps outlined #206 and then ran the tests using `raco test scribble-test` and saw them all pass with racket version v7.9.0.22 [cs] on Ubuntu 20.04


sorawee:
While I agree with the renaming in general, please keep `blueboxes.rkt` because other programs might still be using it! For example, @greghendershott's Racket Mode [requires it](https://github.com/greghendershott/racket-mode/blob/master/racket/scribble.rkt#L15), so this change as it is would break his program.

Similarly, I'm not sure if renaming the Scribble file name is a good idea. Someone might link to https://docs.racket-lang.org/scribble/blueboxes.html literally, and this change would break the link.

scribble-doc/scribblings/scribble/signature-box.scrbl
```diff
@@ -0,0 +1,59 @@
...
+
+@defproc[(make-signature-box-cache
+          [populate? boolean?]
+          [#:signature-box-dirs signature-box-dirs (listof path?) (get-doc-search-dirs)])
+         signature-box-cache?]{
+  Constructs a new (mutable) signature-box cache.
+  
+  If @racket[populate?] is @racket[#f], the cache is initially
+  unpopulated, in which case it is filled in the first time the cache
+  is passed to @racket[fetch-bluebxoes-strs]. Otherwise, the cache is
```
sorawee, line 48: This is an existing typo that you might want to fix.

rfindler:
I agree with @sorawee -- we probably don't want to change any of the existing files or exported names. I think it makes sense to change the documentation to call them "signature boxes" and it may make sense (but is less obviously good) to add new files that just reexport the existing functions under new names (or to rename the existing files and then add backwards compatibility files).

As for the error message -- I think you probably want to work with a more recent version of Racket before trying to investigate that too deeply. Assuming you have build-essentials installed (and maybe a few other packages) you should be able to simply do `git clone https://github.com/racket/racket.git; cd racket ; make` (get coffee) and then `mkdir extra-pkgs ; cd extra-pkgs ; raco pkg update --clone scribble` and then you should be able to work with the right code. hth.

scribble-doc/scribblings/scribble/signature-box.scrbl
nloadholtes: Good catch! Thanks, I'll make sure to get that

nloadholtes:
Great point @sorawee about the external dependencies, I will re-work this PR to prevent breakages.

And thanks @rfindler about the version, as soon as I typed in that used `apt-get` I thought "That is probably the worst way to install racket for this work..." :laughing:

shhyou:
This tutorial by Ben may help: [Tutorial: Contributing to Racket](https://blog.racket-lang.org/2017/09/tutorial-contributing-to-racket.html)

nloadholtes:
Thanks to @shhyou and @rfindler I built racket and re-tested. The tests pass now. I'll turn this PR into a draft PR while I work on the changes requested.

greghendershott:
Not to talk past the close, but, changing the required module name and function names would break more things than just Racket Mode.

- `xrepl` for sure: https://github.com/racket/xrepl/blob/334f40aa3c3cf8616eaf05349c4d3ec99501ced3/xrepl-lib/xrepl/xrepl.rkt#L686

- I vaguely recall at least one `raco` command... but I can't think of it exactly right now.

---

I totally agree that "bluebox" has been kind of an obscure term ever since the docs got a new style sheet, years ago. :smile:

So I definitely think there's merit to changing the docs to help. Like maybe define "bluebox" as a `tech` term, and/or add both "bluebox" and "signature-box" as index terms, and so on.

greghendershott:
Dang I can't recall the `raco` command. Maybe I imagined that.

Anyway, here's one list:

  https://github.com/search?p=2&q=blueboxes-cache&type=Code

That list isn't perfect.

It's too long because it includes many forked repos (although some of those forks might need to be changed, eventually).

It's too short because it's only code on GitHub.

But generally speaking Racket "never" removes modules or files, and breaks existing programs. It might add new, preferred alternatives, or aliases.

nloadholtes:
Thanks @greghendershott! I had no idea this code could have such far reaching impact. :smile: I appreciate all the pointers on this from everyone.

PR #287, opened 2021-01-09
documenting forms: disallow empty box

sorawee:
Forms like `deftogether` and `defproc*` currently allow empty spec,
causing an empty box to be rendered. These malformed outputs are better
prevented at compile-time.

`defform*` already disallows empty spec, but the error is reported
in terms of `defform*/subs`, so this PR adjusts it to directly error
in terms of `defform*`. `defsubform*` still unfortunately errors
in terms of `defform*`, but the fix is too invasive to be worth it.

This PR is tested against all docs in the main distribution.
There's no error.

### DIFF
--- a/scribble-doc/scribblings/scribble/manual.scrbl
+++ b/scribble-doc/scribblings/scribble/manual.scrbl
@@ -1993,7 +1993,7 @@ Returns @racket[#t] if @racket[v] is a bibliography entry created by
 @; ------------------------------------------------------------------------
 @section{Version History}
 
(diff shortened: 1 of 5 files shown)


sorawee:
The latest commit documents the previously undocumented `defthing*` and adjusts `defthing*` to disallow empty box.

Here's the [rendered doc](https://github.com/racket/scribble/files/5791209/4.2.3.Documenting.Forms.Functions.Structure.Types.and.Values.pdf).

PR #290, opened 2021-01-28
Migrate from #lang scheme/base to #lang racket/base

jackfirth:
Large-scale cleanup. No other changes were made to any files.

### DIFF
--- a/scribble-lib/help/help.rkt
+++ b/scribble-lib/help/help.rkt
@@ -1,4 +1,4 @@
-#lang scheme/base
+#lang racket/base
 
 (require "search.rkt" scheme/cmdline scheme/list scheme/string
          raco/command-name) 
--- a/scribble-lib/scribble/base-render.rkt
+++ b/scribble-lib/scribble/base-render.rkt
@@ -1,4 +1,4 @@
-#lang scheme/base
+#lang racket/base
 
 (require "core.rkt"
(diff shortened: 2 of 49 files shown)


scribble-lib/help/help.rkt
```diff
@@ -1,4 +1,4 @@
-#lang scheme/base
+#lang racket/base
 
 (require "search.rkt" scheme/cmdline scheme/list scheme/string
```
sorawee, line 3: Since you are touching these files already, I think it's worth converting these libraries (`scheme/cmdline`, `scheme/list`, etc.) to the corresponding `racket` libraries too.

jackfirth: Agreed, but I want to do that in a separate pull request so it's easier to make sure this first step doesn't break anything.

mflatt:
I get an error when I try to build with this:
```
raco setup: error: during making for <pkgs>/games
raco setup:   scribble-lib/scribble/private/lp.rkt:83:23: all-from-out: no corresponding require
raco setup:     at: scheme/base
raco setup:     in: (all-from-out scheme/base scribble/manual)
raco setup:     compiling: <pkgs>/scribble-lib/scribble/private/lp.rkt
```

jackfirth:
Fixed that error. (Maybe I should set up CI for this repo...)

PR #291, opened 2021-01-28
Cleanup scheme modules

jackfirth:
This pull request builds on #290 by replacing all uses of `scheme/*` modules with their equivalent `racket/*` modules.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,36 @@
+on: [push, pull_request]
+name: CI
+jobs:
+  build:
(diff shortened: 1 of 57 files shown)


mflatt:
This builds fine for me when I merge manually to resolve the conflict. Please rebase (now that the other PR is merged). I'm inclined to squash, since a single renaming event seems more useful to a historical review than individual events, but I'll merge without squashing if you prefer.

jackfirth:
Merged instead of rebasing since that's what was easiest to do with the github web UI. Merging this PR as a squashed commit is good with me: the pull request merge button has squashing as an option right?

PR #292, opened 2021-02-01
Attempt to add CI for scribble packages

jackfirth:
Based on the [Typed Racket CI config](https://github.com/racket/typed-racket/blob/master/.github/workflows/ci.yml). No idea if this works, since I don't have a way to test it.


jackfirth:
Was able to run it successfully on my fork: https://github.com/jackfirth/scribble/runs/1805021750?check_suite_focus=true

samth:
We should really have an abstraction somewhere for that hackery, but I've never been sure of the right place.

PR #293, opened 2021-02-09
Stop using legacy contract forms

jackfirth:
* `false/c` is equivalent to `#f`.
* ~~`flat-contract-predicate` is unnecessary, flat contracts are already predicates.~~


scribble-lib/scribble/decode.rkt
```diff
@@ -84,7 +84,7 @@
 
 (define (spliceof c)
   (define name `(spliceof ,(contract-name c)))
-  (define p (flat-contract-predicate c))
+  (define p c)
```
sorawee, line 87: Perhaps remove `p` entirely?

jackfirth: Done.

scribble-lib/scribble/decode.rkt
```diff
@@ -84,11 +84,10 @@
 
 (define (spliceof c)
   (define name `(spliceof ,(contract-name c)))
-  (define p (flat-contract-predicate c))
   (make-flat-contract #:name name
                       #:first-order (lambda (x)
                                       (and (splice? x)
-                                           (andmap p (splice-run x))))))
+                                           (andmap c (splice-run x))))))
```
LiberalArtist, line 90: I think this change isn't correct: *most* flat contracts are already predicates, but ordinary values that can be promoted to flat contracts (e.g. `'apple`, `#f`, or `#px"\\S"`) can't be used as predicates until they go through are explicitly or implicitly coerced. I think it would be more correct to use `coerce-flat-contract` (and to do it before `contract-name` for error checking, though the `provide/contract` should take care of that).

(Something I don't know: might using `flat-contract-predicate` allow unnecessary parts of the contract to be GC'ed?)

So I think a better change would be something like:
```racket
(define (spliceof c)
  (let* ([c (coerce-flat-contract 'spliceof c)]
         [name `(spliceof ,(contract-name c))])
    (make-flat-contract #:name name
                        #:first-order (lambda (x)
                                        (and (splice? x)
                                             (andmap c (splice-run x)))))))
```

(If I were touching this code, I'd probably make `spliceof` a struct with `prop:flat-contract`, instead.)

jackfirth: Opened https://github.com/jackfirth/resyntax/issues/43, when I get a moment I'll revert this change.

jackfirth: Fixed.

PR #294, opened 2021-02-09
Migrate from let forms to definitions

jackfirth:
This pull request was generated by the automated refactoring tool [`resyntax`](https://github.com/jackfirth/resyntax). It replaces uses of `let` forms with `define` forms, when doing so would decrease nesting and wouldn't change the existing binding structure. Some replacements were not made because `resyntax` isn't always able to preserve comments in the original code.

This is a larger-scale version of #289. While that pull request touched a single file, this one attempts to clean up the entire repository. I can split it into smaller, easier-to-review chunks if preferred. Each file's changes are independent and can be reverted independently, so it should hopefully be easy to undo any mistakes here.

### DIFF
--- a/scribble-lib/help/search.rkt
+++ b/scribble-lib/help/search.rkt
@@ -40,8 +40,8 @@
                                            (format "q?~a" query)))
                                          null))])))]
     [else
-     (define path (build-path (find-user-doc-dir) sub))
-     (let ([path (if (file-exists? path) path (build-path (find-doc-dir) sub))])
+     (let* ([path (build-path (find-user-doc-dir) sub)]
+            [path (if (file-exists? path) path (build-path (find-doc-dir) sub))])
        (notify path)
        (cond
          [(file-exists? path)
--- a/scribble-lib/help/search.rkt
+++ b/scribble-lib/help/search.rkt
@@ -17,38 +17,40 @@
                         #:notify [notify void])
   (define open-url (get-doc-open-url))
   (cond
-   [open-url
-    (define dest-url (let ([u (string->url open-url)])
-                       (combine-url/relative
-                        u
-                        (string-join
-                         (for/list ([s (explode-path sub)])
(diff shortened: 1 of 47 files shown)


scribble-lib/scribble/lp/lang/common.rkt
```diff
@@ -41,16 +41,18 @@
...
-                      (list (shift expr))))
-                (let ([subs (syntax->list expr)])
-                  (if subs
-                      (list (restore expr (loop subs)))
-                      (list (shift expr))))))
+            (cond
+              [(identifier? expr)
+               (define subs (get-chunk expr))
+               (if (pair? subs)
+                   (begin (set! chunk-mentions (cons expr chunk-mentions))
```
sorawee, line 48: Might be out of scope of this PR, but this probably should be converted to `cond`.

jackfirth: Yup, I'll follow up on this with a separate pass that `cond`-ifies more things. I turned off all rules except the let-to-definitions rules when generating this pull request.

scribble-lib/help/search.rkt
```diff
@@ -17,38 +17,40 @@
...
+                                           (url-fragment dest-url))]
+                             [query (append
+                                     (url-query dest-url)
+                                     (if query
+                                         (url-query
+                                          (string->url
+                                           (format "q?~a" query)))
+                                         null))])))]
     [else
-      (let* ([path (build-path (find-user-doc-dir) sub)]
```
sorawee, line 43: This one should not be converted (as discussed in Slack).

jackfirth: Done. Let me know if you see any other instances of this.

jackfirth:
Closing this for now because of jackfirth/resyntax#58. Will generate a new pull request after tweaking `resyntax` a bit.

PR #295, opened 2021-02-12
Replace uses of define-struct with struct

jackfirth:
The `define-struct` form is a legacy form, `struct` is preferred in new code.

Note: these changes were automatically generated by [`resyntax`](https://github.com/jackfirth/resyntax).

### DIFF
--- a/scribble-lib/scribble/base-render.rkt
+++ b/scribble-lib/scribble/base-render.rkt
@@ -1197,8 +1197,9 @@
 
 ;; ----------------------------------------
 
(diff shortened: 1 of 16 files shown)


samth:
I'm a bit skeptical that this is worth it since you are adding `#:extra-constructor-name` in every case. Is it possible to figure out which of those can be omitted?

jackfirth:
I think that ought to be possible, if trickier. It would require making more than one analysis pass over each module. I'll mark this as a draft while I attempt to figure it out.

PR #297, opened 2021-02-23
Make navbar stick and scrollable

wilbowma:
This helps with navigation of long docs pages. For an example, see:
https://www.students.cs.ubc.ca/~cs-411/docs/reference/sets.html?q=sets

### DIFF
--- a/scribble-lib/scribble/manual-style.css
+++ b/scribble-lib/scribble/manual-style.css
@@ -194,7 +194,6 @@ a:hover {
     position: fixed;
     z-index: 1;
     background: #a7b0be;
-    height: auto;
     top: 0;
     left: 0;
     margin-bottom: 0;
--- a/scribble-lib/scribble/manual-style.css
+++ b/scribble-lib/scribble/manual-style.css
@@ -428,9 +428,9 @@ a:hover {
     left: 0;
     top: 0rem;
     width: 14rem;
-    padding: 7rem 0.5rem 0.5rem 0.5rem;
+    padding: 0rem 0.5rem 0.5rem 0.5rem;
     background-color: hsl(216, 15%, 70%);
-    margin: 0;
+    margin: 6rem 0 0 0;
 
 }
 
--- a/scribble-lib/scribble/manual-style.css
+++ b/scribble-lib/scribble/manual-style.css
@@ -191,7 +191,10 @@ a:hover {
 }
 
 .navsettop {
-    position: absolute;
+    position: fixed;
+    z-index: 1;
+    background: #a7b0be;
+    height: auto;
     top: 0;
     left: 0;
     margin-bottom: 0;
@@ -418,7 +421,9 @@ a:hover {
 /* Table of contents, left margin */
 
 .tocset {
-    position: absolute;
+    position: fixed;
+    overflow-y: scroll;
+    height: 88%;
     float: none;
     left: 0;
     top: 0rem;
--- a/scribble-lib/scribble/manual-style.css
+++ b/scribble-lib/scribble/manual-style.css
@@ -423,10 +423,10 @@ a:hover {
 .tocset {
     position: fixed;
     overflow-y: scroll;
-    height: 88%;
     float: none;
     left: 0;
     top: 0rem;
+    bottom: 0;
     width: 14rem;
     padding: 0rem 0.5rem 0.5rem 0.5rem;
     background-color: hsl(216, 15%, 70%);


mflatt:
In my browser (Firefox on Mac OS), I can't scroll all the way to the bottom. By making the `88%` smaller, I can scroll to the end. Is there a replacement for `88%` that will make this work more?

wilbowma:
I started with `80%`, which worked on but was ugly. It's a magic number and if anyone knows more about CSS than me, I'm sure there's a better way.

bhrgunatha:
I like this idea! 

I'm using Firefox 86.0 on linux and I have my browser configured with a minimum font size of 18px so the text may look different but I don't think that affects the scrollbar visibility.

The scroll bar appears to be applied to the full container of the navigation at the left side which means for longer lists it's initially hidden [behind the search form at the top of the navigation section](https://i.imgur.com/aErU2V8.png).

Once you scroll the content list somewhat [it suddenly appears](https://i.imgur.com/dlYVyAN.png)

I think the scrollbar needs to be applied to the list of contents, not the full left hand side.

mydoghasworms:
@wilbowma Well done; this is just as [I hoped it would be](https://www.reddit.com/r/Racket/comments/lqqxt2/a_thought_about_the_racket_docs_having_a_table_of/). Hopefully your change gets merged.

wilbowma:
> @wilbowma Well done; this is just as [I hoped it would be](https://www.reddit.com/r/Racket/comments/lqqxt2/a_thought_about_the_racket_docs_having_a_table_of/).
> 
> Of course, it would be nice for it to be responsive, i.e. to hide the TOC on smaller displays like mobile devices.
> 
> Hopefully your change gets merged and/or someone can make the necessary responsive fixes.

The TOC now hides itself on the phone/tablet display; that was just an error in the deployed example.

> I like this idea!
> 
> I'm using Firefox 86.0 on linux and I have my browser configured with a minimum font size of 18px so the text may look different but I don't think that affects the scrollbar visibility.
> 
> The scroll bar appears to be applied to the full container of the navigation at the left side which means for longer lists it's initially hidden [behind the search form at the top of the navigation section](https://i.imgur.com/aErU2V8.png).
> 
> Once you scroll the content list somewhat [it suddenly appears](https://i.imgur.com/dlYVyAN.png)
> 
> I think the scrollbar needs to be applied to the list of contents, not the full left hand side.

Fixed this I think.

wilbowma:
> In my browser (Firefox on Mac OS), I can't scroll all the way to the bottom. By making the `88%` smaller, I can scroll to the end. Is there a replacement for `88%` that will make this work more?

Okay, a little Googling suggests to set the `top` and `bottom` to 0 rather than the height. It looks better now for me and without magic numbers. Does this fix it for you?

mflatt:
This works well for me!

rfindler:
This works great for me in Safari (I checked "responsive design mode" which shows a lot of differently sized browsers and they all work).

In Chrome it is like the old way, where the table of contents isn't scrollable.

wilbowma:
> In Chrome it is like the old way, where the table of contents isn't scrollable.
Can't reproduce. https://www.students.cs.ubc.ca/~cs-411/docs/reference/sets.html?q=sets looks right to me on google-chrome-88.0.4324.182-1  chromium-88.0.4324.182-1, on linux.

rfindler:
This is with an up-to-date version of chrome under mac os.

https://user-images.githubusercontent.com/399024/109199603-42bcbe00-7765-11eb-85d9-eac6d96f848d.mov

rfindler:
It works like this on safari (same machine) and I think this is how it is supposed to look.


https://user-images.githubusercontent.com/399024/109209957-cda3b580-7771-11eb-9358-d642253ae103.mov

wilbowma:
Very strange. I'll see about breaking on my macOS machine and debugging. What version macOS?

rfindler:
I'm using Mojave (10.14.6).

wilbowma:
Okay I just tried this on 10.14.6 with freshly installed Chromium, and it works as expected. Do you for some reason have an old version of my website cached in Chrome?

rfindler:
oh! I should have thought of that. It works now..... sorry about that.

wilbowma:
No worries. I should have thought of it earlier, but I couldn't imagine why you'd have reason to visit my class site.

rfindler:
Looks like this doesn't always work. For example, with this scribble code in `x.scrbl`:
```
#lang scribble/manual
@title{Title}
```
and this command line:
```
scribble --dest x --html x.scrbl
```
I see this output:

Screen Shot 2021-03-07 at 3 36 22 PM


but in 8.0 I see this output:

Screen Shot 2021-03-07 at 3 35 51 PM

The little white rectangle should probably be grey. Maybe this is because something needs tweaking for the case where the search doesn't appear on the page?

wilbowma:
In `8.0.0.9` I see the expected output; will continue debugging

PR #308, opened 2021-07-18
Fix contracts on Secref and seclink

lexi-lambda:
Both these functions allow the `#:doc` argument to be `#f`, and in fact that is the default value, but their contracts are overly strict, so passing it explicitly triggers a contract violation. This commit therefore weakens the contracts to allow `#f`, as they should.


mflatt:
Thanks!

PR #312, opened 2021-08-27
defsubidform and its document

dannypsnl:


sorawee:
It looks good to me (unless you want to add `maybe-kind maybe-link` to make it match `defidform`).

dannypsnl:
> It looks good to me (unless you want to add `maybe-kind maybe-link` to make it match `defidform`).

I think current one is ok for me

dannypsnl:
remind, does any problem need to resolve here?

mflatt:
Sorry for the delay, I was just looking at this now and about to merge. (It needs a version bump and history note, but I can add that after.)

Did you close because there didn't seem to be activity, or some other reason?

dannypsnl:
Yes, it didn’t seem to be activity :)

mflatt:
Ok! Sorry again for the delay, but I'll merge now.

Issue #313, opened 2021-09-05
Add button to toggle TOC column on low-width devices


jryans:
Overall, I really enjoy reading documents written with Scribble. 🙂 

The one missing thing for me at the moment is access to the TOC when reading on a low-width device. At the moment, the TOC is hidden with no way to access it.

One path forward would be to add a three lines / hamburger icon somewhere on the page that would allow revealing the TOC when desired, perhaps similar to this version used on [Rust docs](https://docs.rs/env_logger/):

<img src="https://user-images.githubusercontent.com/279572/132110853-32c8fd35-4f46-45d2-b4f6-6af11015a962.png" alt="Screen Shot 2021-09-05 at 01 07 00" width="414">

If other people agree this would be useful, I'm happy to send a PR to implement it. 😄

sorawee:
IMO, this would be very useful! A PR is always welcome :)

Issue #305, opened 2021-06-21
Code content can overlap with sidebar


sorawee:
<img width="727" alt="Screen Shot 2021-06-21 at 9 33 33 AM" src="https://user-images.githubusercontent.com/9099577/122796900-c0a45800-d273-11eb-9147-8535fcff1239.png">

sorawee:
To be clear about how to cause this issue:

1. Zoom in
2. Scroll to the left on a `defproc` box

PR #314, opened 2021-09-12
Add table of contents toggle for low-width devices

jryans:
This adds a button to show / hide the TOC column on low-width devices.

It also adjusts various styles to keep the top bar fixed as you scroll so that the TOC toggle button remains accessible anywhere in the document. Flexbox is used in the low-width nav bars to improve vertical centering.

The appearance on high-width devices is unchanged. No TOC button is shown there.

https://user-images.githubusercontent.com/279572/132976575-bc61329e-c1fc-4cb1-a936-4fd9bbcd869d.mov

Fixes https://github.com/racket/scribble/issues/313
Also fixes https://github.com/racket/scribble/issues/305

### DIFF
--- a/scribble-lib/scribble/html-render.rkt
+++ b/scribble-lib/scribble/html-render.rkt
@@ -1017,6 +1017,15 @@
                   (url->string* (path->url up-path))
                   "../index.html")
          `[onclick . ,(format "return GotoPLTRoot(\"~a\");" (version))]))
+      (define tocset-toggle
+        (make-element
+         (make-style "tocsettoggle"
+                     (list
+                      (make-target-url "javascript:void(0);")
+                      (make-attributes
+                       '([title . "show/hide table of contents"]
+                         [onclick . "TocsetToggle();"]))))
+         "contents"))
       (define navleft
         `(span ([class "navleft"])
            ,@(if search-box?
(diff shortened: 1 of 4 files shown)


sorawee:
This is very nice! And it also avoids the problem in #305!

jryans:
> And it also avoids the problem in #305!

Ah yeah, indeed, the `z-index` tweaks I made while overlaying the TOC column also resolve that issue as well. 🙂

mflatt:
Checking before merging: Would it be better to use "TOC" instead of "toc", since it's an abbreviation?

jryans:
@mflatt Thanks for taking a look! 🙂 I wasn’t quite sure what label would be best… Mainly I felt a short lowercase label would fit in best next to top, up, etc., but you’re right that “TOC” is more grammatically correct… Perhaps “nav” (as a short form of navigation) could work as an alternative?

jryans:
It does seem like there’s enough room for “contents” as well, though it’s a tad longer than the other navigation actions.

mflatt:
I don't have a strong opinion, and maybe someone else will have a good idea. But if "contents" fits on relevant screen sizes, then I lean toward that.

jryans:
@mflatt Okay, I have updated to use "contents". It appears to fit down to the iPhone SE size (320 px). Even smaller devices do exist, but the current overall document style needs some tweaking for such small sizes (you start to get horizontal scrolling of the whole page etc.), so I think this feature is okay in assuming "at least 320 px".

mflatt:
Thanks!

PR #328, opened 2022-05-27
Add kind information to the index

sorawee:
This PR creates a new index description `exported-index-desc*` that contain the kind information.
This allows e.g. the search page to display or refine the query
based on the information.

To maintain backward compatibility, the kind information is not added to
any existing structs. Instead, a new index type is created.

The contract for the kind field is `(listof string? (list/c 'code string?))`.
This rich encoding allows typesetting search results like https://docs.racket-lang.org/search/index.html?q=get-admin

Also bump the version, to be used for `#:version` in other `info.rkt`

Original PR description is preserved here:

```
This PR creates new index types that contain the kind information.
This allows e.g. the search page to display or refine the query
based on the information.

Only `form-index-desc`, `procedure-index-desc`, and `thing-index-desc`
are considered, because they are the only ones that currently
allow kind customization. Other indices are already too specific
(e.g. `mixin`, `struct`), so there seems to be no point to allow
customization in such cases. On the other hand, newer forms
that need the kind customization probably should just use
`thing-index-desc(*)`, similar to how the Rhombus documentation
is currently doing.

To maintain backward compatability, the kind information is not added to
any existing structs. Instead, new index types are created at the leaf
level of the index description hierarchy.

It's unclear what the contract of the kind information should be.
Currently, it is `string?`. Here are some other possibilities

- `(or/c string? #f)`: this allows the information to be left out.
  However, I think that anything indexed with form, proc, or thing
  should mandate the kind information, because they default to
  "syntax", "procedure", or "value", so they are already non-empty.
  If one wishes to leave the information out, they can use
  `exported-index-desc` directly (this is what the `syntax/parse`
  library does).

- `xexpr?`: this allows richer encoding. In particular,
  `method-index-desc` currently displays information like
  "get-admin (method of editor<%>)" on the search page.
  This "(method of editor<%>)" is ad-hoc inserted and
  "editor%" is typeseted differently than "method of".
  The idea of `xexpr?` is to make it general enough to typeset
  "(method of editor<%>)" under one single framework,
  without special-casing `method-index-desc` and also allow
  this flexible customization for non-methods.

  On the other hand, allowing `xexpr?` could introduce vulnerabilities
  like HTML injection. We could constrain it, but that seems
  like too much work. So I think continuing to
  special-casing `method-index-desc` might be better.

Therefore, I'm sticking with `string?` for now, until we figure out a
better solution. However, it should be noted that further modification
will risk breaking backward compatibility.

Also bump the version, to be used for `#:version` in other `info.rkt`
```

### DIFF
--- a/scribble-doc/scribblings/scribble/core.scrbl
+++ b/scribble-doc/scribblings/scribble/core.scrbl
@@ -1085,11 +1085,10 @@ The @racket[entry-seq] list must have the same length as
 final document.
 
 The @racket[desc] field provides additional information about the
-index entry as supplied by the entry creator. For example, a reference
-to a procedure binding can be recognized when @racket[desc] is an
-instance of @racket[procedure-index-desc]. See
-@racketmodname[scribble/manual-struct] for other typical types of
-@racket[desc] values.
+index entry as supplied by the entry creator. For example, a reference to
+a procedure binding can be recognized when @racket[desc] is an instance of
+@racket[exported-index-desc*] with the @racket['("procedure")] kind.
+See @racketmodname[scribble/manual-struct] for other types of @racket[desc] values.
 
 See also @racket[index].}
 
--- a/scribble-doc/scribblings/scribble/manual.scrbl
+++ b/scribble-doc/scribblings/scribble/manual.scrbl
@@ -2190,15 +2190,27 @@ exhaustive, because new modules can re-export the binding).}
 Indicates that the index entry corresponds to the definition of a
 syntactic form via @racket[defform] and company.}
 
+@defstruct[(form-index-desc* form-index-desc) ([kind string?])]{
+
+Like @racket[form-index-desc], but adds the kind information.}
+
 @defstruct[(procedure-index-desc exported-index-desc) ()]{
 
 Indicates that the index entry corresponds to the definition of a
 procedure binding via @racket[defproc] and company.}
 
+@defstruct[(procedure-index-desc* exported-index-desc) ([kind string?])]{
+
+Like @racket[procedure-index-desc], but adds the kind information.}
+
 @defstruct[(thing-index-desc exported-index-desc) ()]{
 
 Indicates that the index entry corresponds to the definition of a
 binding via @racket[defthing] and company.}
+
+@defstruct[(thing-index-desc* exported-index-desc) ([kind string?])]{
+
+Like @racket[thing-index-desc], but adds the kind information.}
 
 @defstruct[(struct-index-desc exported-index-desc) ()]{
 
--- a/scribble-doc/scribblings/scribble/manual.scrbl
+++ b/scribble-doc/scribblings/scribble/manual.scrbl
@@ -2185,55 +2185,74 @@ correspond to the documented name of the binding and the primary
 modules that export the documented name (but this list is not
 exhaustive, because new modules can re-export the binding).}
 
+@defstruct[(exported-index-desc* exported-index-desc) ([kind (listof (or/c string? (list/c 'code string?)))])]{
+
+Like @racket[exported-index-desc], but with the @racket[kind] information.
+Common values for @racket[kind] include:
+@racket['("procedure")], @racket['("syntax")], @racket['("value")].
+For the purpose of typesetting, the strings in @racket[kind] will be
+concatenated together, but those tagged with @racket['code]
+will be typeseted as code.
+
+The @racket[exported-index-desc*] struct is preferable over
+other index entry descriptions below.
+
+@history[#:added "1.46"]}
+
 @defstruct[(form-index-desc exported-index-desc) ()]{
+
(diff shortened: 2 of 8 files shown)


samth:
How about `(or/c string? (listof (or/c string? <something that indicates a link>)))`?

samth:
How does this fit with the `index` function?

sorawee:
I didn't modify that at all, so they are completely unaffected.

An `index-element` consists of `tag`, `plain-seq`, `entry-seq`, and `desc`. This PR does not modify this structure, but it creates a new kind of `desc` to replace `form-index-desc`, `procedure-index-desc`, and `thing-index-desc`. Other `desc`s like `tech` (produced by `deftech`), `(part-index-desc)` (produced by `section`), and `#f` (probably produced by `index`) are left untouched.

sorawee:
How about `(define kind/c (listof (or/c string? (list/c 'code string?))))`?

I don't want to `or/c` with `string?` at the outermost level. It's already too complicated.

sorawee:
Oh, another possibility is to create `kind-index-desc` and replace all existing `form-index-desc`, `procedure-index-desc`, and `thing-index-desc` with `kind-index-desc`. That way, we only need to create one more struct, not three.

samth:
I think if we're now using the `desc` in more ways, we want to provide control over it from functions like `index`.

sorawee:
Are you thinking about making the below search result show more information?

Screen Shot 2022-05-27 at 4 44 54 PM
 
I agree that could be helpful, but my opinion is that it is out of scope for this PR. This current PR is already large. I'd deal with `index` / `deftech` / `section` / etc. separately.

sorawee:
By the way, I switched to `exported-index-desc*` as I mentioned above. I think it's an improvement, but it also lost some useful information (though this information is currently not utilized anyway). Let me know if you have a better idea.

sorawee:
It is already the case that `kind` can be any `content?` according to the doc of [`defproc`](https://docs.racket-lang.org/scribble/doc-forms.html#%28form._%28%28lib._scribble%2Fmanual..rkt%29._defproc%29%29), so restricting it to `string?` or `(listof (or/c string? (list/c 'code string?)))` doesn't seem to be a good idea.

samth:
I'm thinking that I'd like the new index entries I added for match (eg app) to mention match in the search results.

sorawee:
Perhaps we want to keep track of three information:

1. `kind`: this is a `content?` that will appear directly in the box.
2. `kind-detail`: this is a `content?` that will appear as a search entry. The intention is that it is as detailed as `kind`, if not more.
3. `kind-id`: probably a string, though a symbol could work as well. The intention is that this will be used programmatically for recognization.

For example, a method would have `kind` being `"method"`, `kind-detail` being `"method of dc%"`, and `kind-id` being `"method"` (or something similar to this).

By default, `kind-detail` is `kind`, and `kind-id` is `(content->string kind)`.

mflatt:
I'd like to pick this thread back up, mainly because I'm interested in adding additional attributes like `language-family` and something to control the order of search results for different kinds.

Since there are clearly many different things that a document might want to communicate to a rendering of the index, I'm thinking that `exported-index-desc*` should add just one new field that is an immutable hash table with symbol keys. A drawback is that the allowed/expected data would depend on the key, and the `exported-index-desc*` constructor would not be able to check values in general, but that seems like an ok trade-off for flexibility.

samth:
1. I agree that a hash table is better than an endless parade of new structs each time we want to add a field.
2. We should explicitly document the known/used keys together with `exported-index-desc*` so that there's a way to generally agree on them.

mflatt:
Yes, I had in mind allowing a mixture of well-known and third-party keys. Would it be better to allow only known keys (in which case the shapes of values associated to keys can be checked)? Or should we allow third-party producers and consumers to agree on additional keys (which are not checked by the `exported-index-desc*` constructor)?

After experimenting with this, I'm less in favor of `'kind` text that is added to the search results as in racket/racket#4263. Specifically, adding "procedure" to most Racket search results seems noisy, and adding "class" after a name that ends `%` seems redundant. The current Rhombus approach of putting the kind in the index entry's text seems to work better; there's a trade-off in letting the producer (document generator) pick versus the consumer (index renderer), and so far it seems like the producer is in a better position to choose whether and how to describe the entry's kind. Adding `'prefix` and `'suffix` might be a useful adjustment of responsibility, though, so that the producer can add text before and after the link, and the consumer can render that distinction.

samth:
I agree about `procedure` and maybe about `class` (although lots of people don't know the convention when searching the docs). But a document-specified kind would be very helpful for types or match expanders and various other things.

mflatt:
Yes, a `'kind-id` as @sorawee  describes seems useful. I've called that just `'kind` in my current draft (PR soon).

sorawee:
Replaced by https://github.com/racket/scribble/pull/442. Closing.

PR #330, opened 2022-06-05
Add "white-space: pre" to RktIn

sorawee:
Whitespaces should be preserved in `litchar`

Before the PR (on page https://docs.racket-lang.org/drracket/Keyboard_Shortcuts.html): 
Screen Shot 2022-06-04 at 5 18 34 PM

After the PR:
Screen Shot 2022-06-04 at 5 18 44 PM


rfindler:
Oh, thank you! I never understood why it looked okay in some browsers and not others!

Issue #333, opened 2022-06-24
`scribble/text-render` doesn't respect `(linebreak)`


LiberalArtist:
Scribble doesn't handle `(linebreak)` properly when generating plain text output. For example:
```
philip@bastet:/tmp/scrbl$ cat scribble-text-newline.scrbl 
#lang scribble/base

abc @(linebreak) 123
philip@bastet:/tmp/scrbl$ scribble --text scribble-text-newline.scrbl 

 [Output to scribble-text-newline.txt]
philip@bastet:/tmp/scrbl$ cat scribble-text-newline.txt 
abc  123
philip@bastet:/tmp/scrbl$
```

It looks like something needs to happen in `render-paragraph`, but I'm not immediately sure how to communicate which newlines to keep—maybe specials?

Related to https://github.com/racket/racket/pull/4301

**Edit:** Fixed the example.

sorawee:
I'm a bit confused. I think you want the _evaluation result_ to be something that has newline characters, not _printing_ a newline character (which is what `newline` does), no?

```
#lang scribble/base

abc @(newline) 123
@(displayln "qwerty")
```

I don't expect to see any `qwerty` in the generated file either (but I do get the newline and `qwerty` in the stdout).

According to the doc of [`decode-flow`](https://docs.racket-lang.org/scribble/decode.html#%28def._%28%28lib._scribble%2Fdecode..rkt%29._decode-flow%29%29):

> A sequence of two or more newlines separated only by whitespace is parsed as a compound-paragraph separator.

So:

```
#lang scribble/base

abc
def
```

is rendered as:

```
abc def
```

while:

```
#lang scribble/base

abc

def
```

is rendered as:

```
abc

def
```

So if you want a new paragraph, perhaps use `'("\n" "\n")` or something like that?

```
#lang scribble/base

abc@|'("\n" "\n")|def
```

for me correctly outputs:

```
abc

def
```

LiberalArtist:
Ah, sorry, I wrote this confusingly. I meant [`(linebreak)`](https://docs.racket-lang.org/scribble/base.html#%28def._%28%28lib._scribble%2Fbase..rkt%29._linebreak%29%29) from `scribble/base`, which produces an element with the style name `'newline`. I'll edit the post above. The implementation in https://github.com/racket/scribble/pull/334 works (with caveats as noted); I'll fix the commit message now.

sorawee:
Ah, yes, I think it is indeed broken. 

[...]

PR #334, opened 2022-06-24
scribble/text-render: support `(linebreak)`

LiberalArtist:
Here's an attempt at an implementation, tested so far only on my current draft of https://github.com/racket/racket/pull/4301.

Related to https://github.com/racket/racket/pull/4301
Closes https://github.com/racket/scribble/issues/333

### DIFF
--- a/scribble-lib/scribble/text-render.rkt
+++ b/scribble-lib/scribble/text-render.rkt
@@ -247,11 +247,33 @@
                  (render-flow d part ht #f)))))))
 
     (define/override (render-paragraph p part ri)
-      (define o (open-output-string))
-      (parameterize ([current-output-port o])
-        (super render-paragraph p part ri))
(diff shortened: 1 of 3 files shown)


LiberalArtist:
I've fixed the commit message here (thanks @sorawee!).

mflatt:
For testing, probably there should be a new document that uses `linebreak` in `scribble/scribble-test/tests/scribble/docs`.

LiberalArtist:
Thanks, I've added a test.

samth:
I don't know this code at all, but is the extra complexity all about line-wrapping? And is the character-at-a-time communication this now has going to be a performance issue?

mflatt:
I don't think this will be a performance issue compared to everything else that goes into running and rendering a document, and I wasn't able to detect a performance change by trying some tests and the build guide.

PR #337, opened 2022-06-28
add `#:indirect?` argument for `tech` and `techlink`

LiberalArtist:
These seem to do the right thing, but the links they generate for https://github.com/racket/racket/pull/4301 don't redirect successfully, and I don't understand why. So, this probably isn't ready to merge yet.

### DIFF
--- a/scribble-doc/scribblings/scribble/core.scrbl
+++ b/scribble-doc/scribblings/scribble/core.scrbl
@@ -1057,7 +1057,13 @@ properties for all @racket[element]s:
        provide an external-link URL, then the resolution of the
        hyperlink can be deferred until the link is clicked (or, in
        some cases, patched by JavaScript when the documentation is
-       viewed in a browser).}
+       viewed in a browser).
+
+       Note that deferred resolution relies on cooperation with the page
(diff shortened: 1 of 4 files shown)


sorawee:
Do you mean to include the `linebreak` change in this PR?

mflatt:
The local-redirect database creation explicitly filters to `'part` and `'mod-path` keys to keep the index smaller:

https://github.com/racket/racket/blob/master/pkgs/racket-index/scribblings/main/private/local-redirect.rkt#L206

I think you'd need to add `'tech` there. Adding `'tech` seems like a reasonable change, but docs.racket-lang.org won't have `'tech` entries until the next release.

LiberalArtist:
Thanks! I was looking at that file, but I hadn't found the right spot.

LiberalArtist:
I've made the corresponding change in a commit I just pushed to https://github.com/racket/racket/pull/4301. The only change here from 8c1dec5 in c9188ec is that I added a note to the documentation for `link-element` explaining that `'indirect-link` does not work with arbitrary tags.

PR #350, opened 2022-09-08
Add Resyntax integration

jackfirth:
This sets up Resyntax analysis on pull requests to this repository. Resyntax will analyze changed files and submit a review of the pull request containing its suggested changes. See [here](https://github.com/racket/typed-racket/pull/1277) for an example.

The workflows are a copy of the workflows I made for Typed Racket. In the future we can make these workflows reusable so they don't have to be copied into every repository, but one step at a time.


jackfirth:
This is ready to merge now.

samth (approved):

PR #364, opened 2023-03-03
Run Resyntax on scribble/private

jackfirth:
This one is a continuation of #363, this time expanded to the entire `scribble/private` collection.

### DIFF
--- a/scribble-lib/scribble/private/manual-bib.rkt
+++ b/scribble-lib/scribble/private/manual-bib.rkt
@@ -21,7 +21,7 @@
                          #:note (or/c #f pre-content?))
              . ->* .
              a-bib-entry?)]
- [rename a-bib-entry? bib-entry? (any/c . -> . boolean?)]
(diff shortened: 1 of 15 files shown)


mflatt:
These improvements look really good — thanks!

mfelleisen:
If resyntax also tore apart 2,000 line syntax functions, it would be perfect :-)

rfindler:
I am a little bit hesitant of the (-> any/c boolean?) to predicate/c
transformation. I feel like the former is easier to read.

Robby

rfindler:
Looks like it does! yay. :) 
```
#lang racket
(struct s ())
(eq? (contract (-> any/c boolean?) s? 'pos 'neg) s?) ;; => #t
```

jackfirth:
Perfect! Then I think it's time to update the docs for `predicate/c` to say it's now an obsolete alias for `(-> any/c boolean?)`, and the latter should just be used directly instead.

PR #367, opened 2023-06-09
tex-table: add guillemetleft and guillemetright

AlexKnauth:
guillemetleft for `«`, and guillemetright for `»`.

See also https://github.com/racket/gui/pull/297

### DIFF
--- a/scribble-lib/scribble/latex-render.rkt
+++ b/scribble-lib/scribble/latex-render.rkt
@@ -1530,6 +1530,8 @@
         [(#\ℜ) "$\\mathfrak{R}$"]
         [(#\ℨ) "$\\mathfrak{Z}$"]
         [(#\ℭ) "$\\mathfrak{C}$"]
+        [(#\«) "\\guillemetleft"]
+        [(#\») "\\guillemetright"]
         [else #f]))
 
     ;; ----------------------------------------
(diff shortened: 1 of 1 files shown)


rfindler:
@AlexKnauth it looks like these changes broke the redex documentation. It looks to me like there needs to be some delimiteres somewhere. In the generated latex for the redex docs I see:

```
\begin{SCodeFlow}\begin{RktBlk}\begin{SingleColumn}\RktPn{(}\RktSym{\badlink{\RktValLink{collection{-}file{-}path}}}\mbox{\hphantom{\Scribtexttt{x}}}\textbf{\guillemotleftfilename.rkt\guillemotright}

\mbox{\hphantom{\Scribtexttt{xxxxxxxxxxxxxxxxxxxxxx}}}\RktVal{"redex"}

\mbox{\hphantom{\Scribtexttt{xxxxxxxxxxxxxxxxxxxxxx}}}\RktVal{"examples"}

\mbox{\hphantom{\Scribtexttt{xxxxxxxxxxxxxxxxxxxxxx}}}\RktVal{"define{-}judgment{-}form"}\RktPn{)}\end{SingleColumn}\end{RktBlk}\end{SCodeFlow}

\noindent replacing \textbf{\guillemotleftfilename.rkt\guillemotright} with one of the names listed above.

```

which looks like it has "filename.rkt" right up against `\guillemotleft` but there should probably be a `{}` between them. 

Looking at the diff here, I think that probably just adding `{}` to the end of the strings should work. I'll give that a try.

AlexKnauth:
Would a similar thing need to be done for `vdots` and `hdots` above that?

rfindler:
> Would a similar thing need to be done for `vdots` and `hdots` above that?

Sounds worth trying to sort out. Maybe the way it happened in the redex docs with guillemets is good place to start when trying to figure that out.

AlexKnauth:
So I found a similar problem with `vdots`. But not with `hdots` because the `hdots` case is dead code, subsumed by the `cdots` case.

PR #368, opened 2023-06-25
tex-table: vdots delimiters, hdots dead

AlexKnauth:
Fixes a problem found in https://github.com/racket/scribble/pull/367#issuecomment-1603679173, of the same kind found for guillemets, but for vdots. Also adds tests for both the guillemet and vdots cases, and deletes the hdots dead code since that's subsumed by the cdots case.


AlexKnauth:
Can this be merged?

PR #377, opened 2023-10-13
feat: add explicit buttons to heading titles

sorawee:
This commit adds two buttons to each heading title.
These buttons are by default hidden, but they will
appear on hovering on the heading titles.

The first button is a "link here" button, which adjusts the URL of the
current page to include an anchor to link to that heading title, similar
to how GitHub makes each heading title in README documents a link
to add an anchor.

The second button is to view Scribble source and internal link.
Previously, clicking heading titles serves this functionality,
but without an indicator that heading titles can be clicked,
the feature is not discoverable. So this commit makes a
transition toward an explicit button.

Example screenshots:

Page title

Screenshot 2023-10-14 at 1 38 13 PM

Section title

Screenshot 2023-10-14 at 1 38 04 PM

### DIFF
--- a/scribble-lib/scribble/manual-racket.js
+++ b/scribble-lib/scribble/manual-racket.js
@@ -236,12 +236,15 @@ function AddPartTitleOnClick(elem) {
         else
             elem.parentNode.appendChild(info);
 
-        /* Clicking the header shows the explanation element: */
-        elem.onclick = function () {
-            if (info.style.display == "none")
+        /* Clicking the information button shows the explanation element: */
+        const heading = elem.querySelector('.heading-source');
+        if (heading) {
+          heading.onclick = function () {
+            if (info.style.display === "none")
                 info.style.display = "block";
             else
                 info.style.display = "none";
+          }
         }
     }
 }
--- a/scribble-lib/scribble/html-render.rkt
+++ b/scribble-lib/scribble/html-render.rkt
@@ -4,6 +4,7 @@
          "html-properties.rkt"
          "private/literal-anchor.rkt"
          racket/class
+         racket/match
          racket/path
          racket/file
          racket/port
@@ -1154,6 +1155,7 @@
                              (append
                               (if (and src taglet)
                                   `([x-source-module ,(format "~s" src)]
+                                    [class "heading"]
                                     ,@(let* ([path (resolved-module-path-name
                                                     (module-path-index-resolve
                                                      (module-path-index-join src #f)))]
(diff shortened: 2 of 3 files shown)


scribble-lib/scribble/manual-racket.js
```diff
@@ -236,12 +236,15 @@ function AddPartTitleOnClick(elem) {
         else
             elem.parentNode.appendChild(info);
 
-        /* Clicking the header shows the explanation element: */
-        elem.onclick = function () {
-            if (info.style.display == "none")
+        /* Clicking the information button shows the explanation element: */
+        const heading = elem.querySelector('.heading-source');
+        if (heading) {
```
samth, line 241: Why is this test needed? Can it ever fail?

sorawee: Technically, yes. If users insert, say, an `h3` element with custom attributes that happen to match the structure that we are looking for, but did not have the corresponding `.heading-source` button, then `heading` would be `null`.

jryans (commented):
Thanks for working on this, I agree it's good to make this more discoverable than it currently is. 😄 

I wonder if it's a bit too visually noisy / distracting to show these buttons all the time though...? Some sites that do this sort of thing only show the buttons for a single header when you hover on the header text. Sure, it reduces discoverability a bit, but you're still bound to encounter them after a bit of reading time. Just a suggestion from my initial reaction to your screenshots.

About vertical alignment, the buttons seem to be a few pixel lower than I'd expect. Here's GitHub vs. this PR (with an added line at the text baseline):

GitHub:

2023-10-13 at 14 25

This PR:

2023-10-13 at 14 27

I would recommend nudging the icons up a few pixels so that none of them drop below the text baseline. If you're not sure how, I could also give it a try if you like.

sorawee:
Still not perfect, but this should be a bit better.

Screenshot 2023-10-13 at 9 13 31 PM

(It can be made better, but that will require a percentage value that is "ugly" instead. This is my best attempt without resorting to using the ugly numbers.)

mflatt:
I'll second @jryans's suggestion and rationale to show the buttons only when mousing over the title.

sorawee:
I made the update as requested.

Since we now show the buttons only on hovering over the heading titles, I removed the adjustments to make the buttons smaller*, to help with discoverability.

[*] well, there's still a place that needs this adjustment, since it would be too large with the full size.

mflatt:
Thanks for the update! This looks good to me.

PR #379, opened 2023-10-14
refactor: modernize JS in scribble-common.js

sorawee:
Browsers released within the last 5 years (the partial support tier according to https://github.com/racket/scribble/pull/240) support URLSearchParams, so we should use it.

### DIFF
--- a/scribble-lib/scribble/scribble-common.js
+++ b/scribble-lib/scribble/scribble-common.js
@@ -3,67 +3,24 @@
 // Page Parameters ------------------------------------------------------------
 
 var page_query_string = location.search.substring(1);
-
-var page_args =
-  ((function(){
-      if (!page_query_string) return [];
-      var args = page_query_string.split(/[&;]/);
-      for (var i=0; i<args.length; i++) {
-        var a = args[i];
-        var p = a.indexOf('=');
-        if (p >= 0) args[i] = [a.substring(0,p), a.substring(p+1)];
-        else args[i] = [a, false];
-      }
-      return args;
-    })());
+const page_args = new URLSearchParams(location.search);
 
 function GetPageArg(key, def) {
-  for (var i=0; i<page_args.length; i++)
-    if (page_args[i][0] === key) {
-      try {
-        return decodeURIComponent(page_args[i][1]);
-      } catch (e) {
-        if (e instanceof URIError) {
-          return page_args[i][1];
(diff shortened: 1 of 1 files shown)


sorawee:
As a side effect, this handles query string much better than what it currently does. For example, the malformed query string in https://github.com/racket/scribble/pull/329 / https://github.com/greghendershott/racket-mode/issues/623 can now be decoded with no issue.

scribble-lib/scribble/scribble-common.js
```diff
@@ -3,67 +3,27 @@
...
-  return def;
+  return page_args.get(key) || def;
 }
 
 function MergePageArgsIntoLink(a) {
-  if (page_args.length == 0 ||
-      (!a.attributes["data-pltdoc"]) || (a.attributes["data-pltdoc"].value == ""))
+  if (page_args.size === 0 ||
+      (!a.attributes["data-pltdoc"]) ||
+      (a.attributes["data-pltdoc"].value === ""))
```
yfzhe, line 15: use `dataset` here? 

```javascript
(!a.dataset.pltdoc || a.dataset.pltdoc === "")
```
or even just
```javascript
!a.dataset.pltdoc
```
because the result of accessing dataset value will only be `undefined` or a string.

scribble-lib/scribble/scribble-common.js
```diff
@@ -3,67 +3,27 @@
...
-        if (args[j] == page_args[i][0]) { exists = true; break; }
-      if (!exists) str += "&" + page_args[i][0] + "=" + page_args[i][1];
-    }
-    return prefix + "?" + str + suffix;
+  const url = new URL(href, window.location.href);
+  for (const [key, val] of page_args) {
+    if (url.searchParams.has(key)) continue;
+    url.searchParams.append(key, val)
+  }
+  return url.toString();
```
yfzhe, line 26: I think `url.href` is better.

scribble-lib/scribble/scribble-common.js
sorawee: Updated.

scribble-lib/scribble/scribble-common.js
sorawee: Updated.

sorawee:
Thank you for the feedback, @yfzhe!

PR #391, opened 2023-12-09
feat: support #:auto-value for defparam and defthing

sorawee:
Manually specifying `#:value` could be cumbersome. An example is `html-empty-tags` from the `xml` collection, which is a list of more than 10 symbols.
It is also error-prone, and could be out-of-sync from the actual value.

This PR adds a support for `#:auto-value` in `defparam` and `defthing` to automatically query the value from the `for-label` binding.

Thanks to @LiberalArtist for the help in https://github.com/racket/racket/pull/4807.

### DIFF
--- a/scribble-doc/scribblings/scribble/manual.scrbl
+++ b/scribble-doc/scribblings/scribble/manual.scrbl
@@ -2,6 +2,7 @@
 @(require scribble/manual "utils.rkt"
           (for-syntax racket/base)
           (for-label scribble/manual-struct
+                     racket/list
                      version/utils
                      syntax/quote))
 
@@ -1308,7 +1309,7 @@ Examples:
 
 @defform[(defparam maybe-link id arg-id
            contract-expr-datum
-           maybe-value
+           maybe-auto-value
            pre-flow ...)]{
 
 Like @racket[defproc], but for a parameter. The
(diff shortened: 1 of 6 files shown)


scribble-lib/scribble/private/manual-proc.rkt
```diff
@@ -140,6 +141,28 @@
...
                        #:defaults ([value #'no-value]))))
+
+ (define-splicing-syntax-class (autoable-value-kw name proc)
+   #:description "#:value or #:auto-value keyword"
+   (pattern (~seq #:value extracted))
+   (pattern (~seq #:auto-value)
+            #:do [(define binding-info (identifier-binding name #f))
+                  (unless (list? binding-info)
+                    (raise-syntax-error #f "not imported for-label" name))
+                  ;; the third result is nominal-from-mod, which is what we want
```
sorawee, line 152: I'm actually not sure about this...

sorawee:
This strategy is not foolproof though. For example, https://github.com/racket/scribble/blob/master/scribble-doc/scribblings/scribble/utils.rkt#L12-L13 bounces the for-label binding, but not the phase 0 binding. So without `(require (for-label racket/base))`, `current-readtable` would be identified as originated from `utils.rkt`, which is not the case. I'm not sure if there's a way to fix the problem. We could also just document the limitation and call it a day...

sorawee:
Switching from the third result to the first result (`nominal-from-mod` -> `from-mod`) seems to help with the above issue.

sorawee:
Addressed feedback by @rocketnia.

scribble-doc/scribblings/scribble/manual.scrbl
```diff
@@ -1379,6 +1392,11 @@ If @racket[#:value value-expr-datum] is given, @racket[value-expr-datum]
 is typeset using @racket[racketblock0] and included in the documentation.
 Wide values are put on a separate line.
 
+@racket[#:auto-value] is similar to @racket[#:value value-expr-datum],
```
mflatt, line 1395: Nit: I'd rephrase to avoid starting the sentence with a lowercase keyword, maybe "Supplying #:auto-value ...".

scribble-test/tests/scribble/docs/manual-ex.rkt
```diff
@@ -20,3 +20,14 @@
 (struct pn (x y))
 
 (define v 10)
+
+(require racket/fixnum
```
mflatt, line 24: Nit: I'd prefer to keep `require`s at the top.

mflatt (approved):
Looks ok to me. Ignore nit-picking comments as you see fit.

sorawee:
Oof, I should have bumped the version and added a history note. Will push a fix.

PR #398, opened 2023-12-19
typeset: correct the form of `code:contract#` and `code:quote`

sorawee:
- `code:contract#` only allows one subform in the implementation, but the documentation allows many subforms.
- `code:quote` only allows one subform in the documentation, but the implementation allows many subforms.

This PR corrects both the implementation and the documentation so that the implementation and the documentation agree with each other.


mfelleisen:
My preference would be to go with one item per form: `(code:quote _datum)` and `(code:contract# _datum)`

PR #400, opened 2023-12-25
Refactor racket.rkt

sorawee:
This PR consists of two parts: semi-automatic fixes by Resyntax, and manual fixes.

### Resyntax fixes

Resyntax fixes are commits with the prefix `resyntax:`. Each commit corresponds to fixes due to a rule. The methodology to produce fixes is as follows:

1. Run Resyntax, and pick "the first fix".
2. Apply all fixes with the same rule as "the first fix". 
3. Re-run Resyntax again to possibly get more fixes, and go to step 3 until we reach the fixed point.

Resyntax is modified to omit some rules that are not suitable for the file, but it is also made more lenient in presence of comments (https://github.com/jackfirth/resyntax/issues/232), making it possible to discover fixes that the standard Resyntax can't discover.

### DIFF
--- a/scribble-lib/scribble/racket.rkt
+++ b/scribble-lib/scribble/racket.rkt
@@ -651,14 +651,14 @@
                (unless no-cons?
                  (out (let ([s (cond
                                  [(pair? (syntax-e c))
-                                  (if (syntax->list c)
-                                      "list"
-                                      (if (let ([d (cdr (syntax-e c))])
-                                            (or (pair? d)
-                                                (and (syntax? d)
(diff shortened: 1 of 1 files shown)


sorawee:
I have more changes to make, but this is a large enough PR already, so I'd rather stop here for now.

mflatt:
LGTM, but if it's not pressing, just after the v8.12 branch? (Not that problems seem likely, but just in case.)

sorawee:
Definitely not pressing!

samth:
Let's merge this now so we don't forget.

PR #401, opened 2024-01-04
Refine the grammar of `field-name` in `defstruct`.

shhyou:
The `defstruct[*]` form does support field options. Therefore use the same grammar as [`[define-]struct`](https://docs.racket-lang.org/reference/define-struct.html#%28form._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._struct%29%29) in the documentation.

### DIFF
--- a/scribble-doc/scribblings/scribble/manual.scrbl
+++ b/scribble-doc/scribblings/scribble/manual.scrbl
@@ -1467,16 +1467,20 @@ Examples:
 
 
 @deftogether[(
-@defform[       (defstruct* maybe-link struct-name ([field-name contract-expr-datum] ...)
+@defform[       (defstruct* maybe-link struct-name ([field contract-expr-datum] ...)
                   maybe-mutable maybe-non-opaque maybe-constructor
                   pre-flow ...)]
-@defform/subs[  (defstruct maybe-link struct-name ([field-name contract-expr-datum] ...)
+@defform/subs[  (defstruct maybe-link struct-name ([field contract-expr-datum] ...)
                   maybe-mutable maybe-non-opaque maybe-constructor
                   pre-flow ...)
                ([maybe-link code:blank
                             (code:line #:link-target? link-target?-expr)]
                 [struct-name id
                              (id super-id)]
+                [field field-id
+                       (field-id field-option ...)]
+                [field-option #:mutable
+                              #:auto]
                 [maybe-mutable code:blank
                                #:mutable]
                 [maybe-non-opaque code:blank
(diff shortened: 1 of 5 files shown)


scribble-doc/scribblings/scribble/manual.scrbl
```diff
@@ -1467,16 +1467,20 @@ Examples:
...
-@defform/subs[  (defstruct maybe-link struct-name ([field-name contract-expr-datum] ...)
+@defform/subs[  (defstruct maybe-link struct-name ([field contract-expr-datum] ...)
                   maybe-mutable maybe-non-opaque maybe-constructor
                   pre-flow ...)
                ([maybe-link code:blank
                             (code:line #:link-target? link-target?-expr)]
                 [struct-name id
                              (id super-id)]
+                [field field-id
+                       (field-id field-option ...)]
```
sorawee, line 1481: In the struct implementation, this is valid:

```
(struct foo ([bar]))
```

However, the same in `defstruct*` is not valid:

```
@defstruct*[foo ([(bar) any/c])]
```

Either the implementation of `defstruct[*]` should be adjusted, or the documentation should change `...` with `...+`.

sorawee: Also, should there be checks to make sure that `field-option` is either `#:mutable` or `#:auto`? The current implementation allows any keyword.

shhyou: Thanks for catching this! I fixed `defstruct` to make the empty list of options work.

For the `field-option` checks, it seems like most similar checks in Scribble are syntax errors. I don't know how to do it so I'll leave that out.

scribble-test/tests/scribble/docs/manual.scrbl
```diff
@@ -99,6 +99,14 @@ A function, again, not a link target, documented to return @racket[10] using a d
 
 @defstruct[#:link-target? #f pt ([x real?] [y real?]) #:mutable]{A mutable structure type with extra name, again.}
 
+@defstruct*[#:link-target? #f counter ([name symbol?] [(count) exact-nonnegative-integer?])]{A struct with an (empty) list of field options.}
+
+@defstruct*[counter ([name symbol?] [(count #:mutable) exact-nonnegative-integer?])]{Another struct with a mutable field.}
+
+@defstruct*[#:link-target? #f counter ([name symbol?] [(count #:auto-value) exact-nonnegative-integer?])]{A struct with an automatic field.}
```
sorawee, line 106: It's called `#:auto`, not `#:auto-value`, FWIW.

shhyou: Fixed. Thanks!

For the `field-option` checks, it seems like most similar checks in Scribble are syntax errors. I don't know how to do it so I'll leave that out.

sorawee:
This looks good to me!

shhyou:
Thanks a lot for reviewing!

I don't have push access. @samth could you help merging?

PR #403, opened 2024-01-06
core: add itemization-ordered style property

sorawee:
`itemization-ordered` makes it possible to start an ordered itemization at an arbitrary number. This style property is supported in all backends.

Also implement ordered itemization for the Markdown and text backends (previously, all itemizations are considered unordered).


sorawee:
Which one is better? 
1. The `itemization-ordered` style property must always be used with the `'ordered` style name for things to work correctly (this PR, though there's no check).
2. The `itemization-ordered` style property overrides the style name to be `'ordered`.

I don't like 1) since it has an illegal representation, but 2) also seems bad. E.g., what if I have a `'compact` style name and `itemization-ordered` style property? What do I do?

 Is there better solutions?

samth:
The scenario you describe seems like it should be an error.

sorawee:
OK, so it sounds like @samth favors (1) but with a check (and I agree).

If I use `itemization-ordered` on non `itemization`, should that cause an error too?

Note that Scribble generally has no check for anything like this at all. For example:

```
#lang scribble/manual

@title{Some title}

@itemlist[
 #:style 'style-name-that-Scribble-doesnt-recognize
 @item{first item}
 @item{second item}
]
```

currently works fine, even though the invalid style name should cause an error. I don't know if they are allowed for forward compatibility or any other reasons.

Issue #411, opened 2024-03-05
[Feature Request] Allow `math` to interpret content within `{}` as subscripts or superscripts


NoahStoryM:
**Is your feature request related to a problem? Please describe.**
When using the `math`, only the non-empty sequences of numbers and letters after
`_` or `^` are interpreted as subscripts or superscripts, which limits the
usability of mathematical expressions to denote subscripts or superscripts.

**Describe the solution you'd like**
I propose enhancing the math interpret the content within `{}` as either
subscripts or superscripts. For instance, in the expression `f^{a+b}_{x-y}`,
`a+b` should be interpreted as a superscript, and `x-y` should be interpreted
as a subscript.

**Do you want to contribute to this feature**
I'm interested in contributing to the development of this feature if needed.

PR #412, opened 2024-03-06
Allow `math` to interpret content within `{}` as a (sub)superscript

NoahStoryM:
See #411.


scribble-lib/scribble/private/manual-style.rkt
```diff
@@ -219,26 +219,55 @@
...
            (cond
-             [(regexp-match #px"^(.*)_([a-zA-Z0-9]+)(.*)$" i)
+             [(regexp-match #px"^(.*)\\^\\{(.*)\\}_\\{(.*)\\}(.*)$" i)
               => (lambda (m)
-                   (append (loop (cadr m))
-                           (list (make-element 'subscript
-                                               (loop (caddr m))))
-                           (loop (cadddr m))))]
-             [(regexp-match #px"^(.*)\\^([a-zA-Z0-9]+)(.*)$" i)
+                   (let* ([m (cdr m)] [s1 (car m)]
```
sorawee, line 224: Using `match` to extract `s1`, ... would be better.

sorawee: Note that `match` already supports a `pregexp` pattern, so you can replace the entire `cond` with a `match` on `i`.

sorawee:
Please also add some tests in https://github.com/racket/scribble/blob/master/scribble-test/tests/scribble/docs/manual.scrbl

NoahStoryM:
I don't know how to test in `manual.scrbl` that `math` interprets superscript and subscript correctly, maybe I should directly test that `pregexp` is correct?

sorawee:
You can write some content using `math` in the Scribble file, and then render it to text (https://github.com/racket/scribble/blob/master/scribble-test/tests/scribble/docs/manual.txt) and see if that's the result that you expect. If so, save the change. This will become the expected result that the test runner will compare against.

scribble-lib/scribble/private/manual-style.rkt
```diff
@@ -218,27 +219,25 @@
...
-             [(regexp-match #px"^(.*)([()0-9{}\\[\\]\u03C0])(.*)$" i)
-              => (lambda (m)
-                   (append (loop (cadr m))
-                           (list (caddr m))
-                           (loop (cadddr m))))]
-             [else
-              (list (make-element 'italic (list i)))])]
-          [(eq? i 'rsquo) (list 'prime)]
+           (match i
+             [(or (pregexp #px"^(.*)(_|\\^)\\{(.*)\\}(.*)$"
```
sorawee, line 223: I'm actually skeptical now that you can do this properly with regexp, since braces are supposed to be balanced. 

Take `A_{B^{C}}` as an example. With the current regexp, you would get `"A_{B"` as the first part and `"C}"` as the second part.

NoahStoryM:
I just realized that we can achieve similar functionality by defining functions:

```racket
(define ^ (compose superscript math))
(define _ (compose subscript math))
```

I'm not sure if this PR is necessary, feel free to close it.

sorawee:
I will close the PR then, thanks.

PR #415, opened 2024-04-10
WIP Improve acmart support

fare:


fare:
TODO
[ ] Complete the documentation update
[ ] Add a regression test of some form
[ ] Build and test

scribble-doc/scribblings/scribble/acmart.scrbl
```diff
@@ -35,9 +44,8 @@ Example:
 @defidform[sigplan]
 @defidform[sigchi]
 @defidform[sigchi-a]
-@defidform[dtrap]
```
sorawee, line 38: Would it be possible to preserve these so that old programs that use them can still be run?

PR #417, opened 2024-04-17
update to v2.07

rfindler:
Shall we also include this in the release?

(Just as a sanity check I rebuilt a recent paper and there were no problems.)


sorawee:
2.07 is just released, too. Let's target that one instead.

rfindler:
Okay, I've updated the pull request. The paper still seems to build.

sorawee:
LGTM! I don't think it is that urgent to be included in the release though.

PR #432, opened 2024-08-22
Run Resyntax on markdown renderer

jackfirth:
This pull request uses `Resyntax` to clean up the `scribble/markdown-render` module. It's using a newer version of Resyntax which also uses `fmt` to format the fixed code. I'm opening this PR to get feedback on whether the formatting changes seem like an improvement.


mflatt:
Thanks! Should I squash these commits to merge, to keep them separate?

jackfirth:
I recommend squashing. They're only separate because Resyntax doesn't do multiple passes at once yet.

PR #442, opened 2024-09-26
Add `exported-index-desc*`

mflatt:
This is a revised version of #328

The `exported-index-desc*` struct adds an `extras` field that is a composable catch-all for extensions. The set of keys in `extras` is meant to be a mixture of well-known keys and third-party keys that a producer and consumer might agree on; the contract on `exported-index-desc*` checks that well-known keys have values of the right shape.

The initial well-known keys are `'kind` (corresponding to the text in the upper right of a documentation box), `'hidden?` (indicates an entry that looks redundant to a user, so it should be filtered out when rendering), `'method`, `'constructor?`, and `class-tag`. Those last 3 cover the information that used to be provided by `method-index-desc` and `constructor-index-desc`.

For a method, the index text now includes `(method of <class>)` text that is otherwise added as a special case for `method-index-desc` in the documentation search panel. The intent here is to put those decisions more consistently in the document producer, not the index consumer. Also, `(struct)` is added to struct type entries, because there's no naming convention that makes that otherwise apparent.

Accompanying change to search generation: https://github.com/mflatt/racket/tree/indexdesc

### DIFF
--- a/scribble-doc/scribblings/scribble/manual.scrbl
+++ b/scribble-doc/scribblings/scribble/manual.scrbl
@@ -2250,43 +2250,111 @@ correspond to the documented name of the binding and the primary
 modules that export the documented name (but this list is not
 exhaustive, because new modules can re-export the binding).}
 
+@defstruct[(exported-index-desc* exported-index-desc) ([extras (hash/dc [k symbol?]
+                                                                        [v (k)
+                                                                           (case k
+                                                                             [(kind) string?]
+                                                                             [(hidden?) boolean?]
+                                                                             [(method-name) symbol?]
+                                                                             [(constructor?) boolean?]
+                                                                             [(class-tag) tag?]
+                                                                             [else any/c])]
+                                                                        #:immutable #t)])]{
+
+Like @racket[exported-index-desc], but with additional optional
+information in an @racket[extras] hash table. All values in the ahsh
+table should be @racket[write]able in the sense that @racket[read]
+will reconstruct the value. Any symbol is allowed as a key in
+@racket[extras], but some are well-known:
+
+@itemlist[
+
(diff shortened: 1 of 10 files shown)


scribble-doc/scribblings/scribble/manual.scrbl
```diff
@@ -2250,44 +2250,112 @@ correspond to the documented name of the binding and the primary
...
+       added to the index entry's main text as shown to a user, but it
+       be rendered by some interfaces, such as to the side or in hover
+       text.}
+
+ @item{@racket['hidden?]: A boolean indicating that the index entry
+       describes a binding that is covered from a user's perspective
+       by a different index entry. These are considered redundant and
+       filtered from index rendering.}
+
+ @item{@racket['method-name]: A symbol indicating that the index entry
```
sorawee, line 2284: This is only applicable when `kind` is `method`, right? It might be nice to document the invariant.

scribble-doc/scribblings/scribble/manual.scrbl
```diff
@@ -2250,44 +2250,112 @@ correspond to the documented name of the binding and the primary
...
+       describes a method for the class that is in the @racket[name]
+       field, and the symbol is the method's name.}
+
+ @item{@racket['constructor?]: A boolean indicating that the index
+       entry describes a class's constructor, separate from an index
+       entry for the class. A constructor normally also has
+       @racket['hidden?] as true.}
+
+ @item{@racket['class-tag]: A tag that links to the class's main entry
```
sorawee, line 2293: Ditto for `'class` `kind`?

sorawee (commented):
A too late review, since it's already merged...

Issue #285, opened 2020-12-03
Dark mode for scribble docs


jackfirth:
With the `prefers-color-scheme` media query, CSS stylesheets can [detect whether the user has dark mode](https://css-tricks.com/dark-modes-with-css/) enabled in their operating system settings. Can we get a dark theme for Scribble docs? I tend to code at night a bit more than I like to admit, and the bright whites of scribble docs cause a lot of eye strain. Paging @mbutterick for typography concerns and general web design wizardry.

mbutterick:
I’m afraid I must decline. I am morally opposed to dark mode. It would not be a difficult project for a true believer.

jackfirth:
Fascinating, and as a subscriber to your tinyletter mailing list I'd love to hear your thoughts about it sometime.

Anyway, noted. Perhaps someday I'll give this a shot.

Terkwood:
I worked around this on OSX by learning a bit more about the accessibility system. 


If you already have the accessibility icon available in your top bar, you can navigate to the Accessibility Preferences choice (also available somewhere in system preferences)

<img width="194" alt="Screen Shot 2021-01-25 at 3 05 26 PM" src="https://user-images.githubusercontent.com/38859656/105760000-27208680-5f1f-11eb-8895-2334b2ea6a35.png">

I made Invert Colors available 

<img width="539" alt="Screen Shot 2021-01-25 at 3 05 46 PM" src="https://user-images.githubusercontent.com/38859656/105760080-40293780-5f1f-11eb-9f03-5ad1f58ae26a.png">


I also searched separately for "shortcuts" and made sure that the keyboard binding  is available, which makes this basically usable 

<img width="635" alt="Screen Shot 2021-01-25 at 3 06 05 PM" src="https://user-images.githubusercontent.com/38859656/105760108-47e8dc00-5f1f-11eb-99c1-e800b82e239d.png">

Terkwood:
Hey I think I misread the above comment from @mbutterick .  Sorry.

Just to clarify, are PRs welcome ?

samth:
Yes, PRs for this (and other things) are certainly welcome.

beelzebielsk:
Where would I look to take a shot at this? Or at least make a stylesheet for personal use?

samth:
The stylesheet for scribble is at https://github.com/racket/scribble/blob/master/scribble-lib/scribble/scribble.css as well other .css files in that directory.

PR #452, opened 2024-10-27
scribble dark-mode

dr-neptune:
Hi folks!

I have taken a crack at making a dark-mode for scribble that is enacted with a media-query fir prefers-color-scheme: dark per this issue: https://github.com/racket/scribble/issues/285

Here are some screenshots (in the real version the sidebar stays constant):

Screenshot 2024-10-26 at 20-59-38 4 10 Pairs and Lists

### DIFF
--- a/scribble-lib/scribble/manual-racket.css
+++ b/scribble-lib/scribble/manual-racket.css
@@ -324,3 +324,120 @@ tbody > tr:first-child > td > .together {
     padding: 0.2rem;
     text-align: left;
 }
+
+@media (prefers-color-scheme: dark) {
+    a.RktValDef, a.RktStxDef, a.RktSymDef,
+    span.RktValDef, span.RktStxDef, span.RktSymDef {
+        color: #b5c2e6;
+
+    }
+
+    .inherited {
+        background-color: #9E1C1F;
+
+    }
+
+    .RktValLink, .RktStxLink, .RktModLink {
+        color: #e6ccff;
+    }
+
+    h2 a.RktStxLink, h3 a.RktStxLink, h4 a.RktStxLink, h5 a.RktStxLink,
+    h2 a.RktValLink, h3 a.RktValLink, h4 a.RktValLink, h5 a.RktValLink,
+    h2 .RktSym, h3 .RktSym, h4 .RktSym, h5 .RktSym,
+    h2 .RktMod, h3 .RktMod, h4 .RktMod, h5 .RktMod,
+    h2 .RktVal, h3 .RktVal, h4 .RktVal, h5 .RktVal,
+    h2 .RktPn, h3 .RktPn, h4 .RktPn, h5 .RktPn {
+        color: #e6ccff;
+
+    }
+
+    .tocset td a.tocviewselflink .RktValLink,
+    .tocset td a.tocviewselflink .RktStxLink,
+    .tocset td a.tocviewselflink .RktMod,
+    .tocset td a.tocviewselflink .RktSym {
+        color: #e6ccff;
+    }
+
+    .toptoclink .RktStxLink, .toclink .RktStxLink,
+    .toptoclink .RktValLink, .toclink .RktValLink,
+    .toptoclink .RktModLink, .toclink .RktModLink {
+        color: #e6ccff;
+
+    }
+
+    .tocset .RktValLink, .tocset .RktStxLink, .tocset .RktModLink, .tocset .RktSym {
+        color: #e6ccff;
+
+    }
+
+    .SVInsetFlow .RktVar {
+        color: #e6ccff;
+
+    }
+
+    .RktVar {
+        color: #b5c2e6;
+
+    }
+
+    .RktRes {
+        color: #ffc9d1;
+    }
+
+    .RktSym {
+        color: #b5c2e6;
+
+    }
+
+    .RktOut {
+        color: #e6ccff;
+    }
+
+    .defmodule {
+        border-top: 1px dotted #99b;
+        background-color: #2c2e35;
+        border-bottom-left-radius: 5px;
+        border-bottom-right-radius: 5px;
+    }
+
+    .defmodule a {
+        color: #e6ccff;
+    }
+
+    .defmodule .RktPn {
+        color: #b35f61;
+    }
+
+    .RktVal {
+        color: #ffc9d1;
+    }
+
+    .RktMeta {
+        color: #e6ccff;
+    }
+
+    .RktPn {
+        color: #b35f61;
+    }
+
+    .RktErr {
+        color: #9E1C1F;
+    }
+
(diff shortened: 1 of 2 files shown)


sorawee:
The theme generally looks really nice! A couple of random thoughts:

I personally prefer more contrast (especially for code, like identifiers).  

I notice the font change. E.g., this is how headings look from my computer, which doesn't look like the screenshot. Does the change happen because your computer doesn't support Fira? Or is it something else?

Screenshot 2024-10-26 at 6 41 24 PM

Do you think it makes sense to change the finger point from an image (https://docs.racket-lang.org/reference/finger.png) to a unicode text, so that the color changes accordingly in the dark mode?

Have you checked documentation pages that display graphics (e.g., the output is a filled black square)? Do they work well?

Do you plan to add an explicit button to switch the theme (probably not as a part of this PR, but in the future)?

dr-neptune:
> The theme generally looks really nice! A couple of random thoughts:
> 
> I personally prefer more contrast (especially for code, like identifiers).
> 

I can put together a higher contrast version and post screenshots here. I also prefer higher contrast, but part of the appeal of lower contrast is that it is less jarring at night-time.


> I notice the font change. E.g., this is how headings look from my computer, which doesn't look like the screenshot. Does the change happen because your computer doesn't support Fira? Or is it something else?
> 

This is because my computer doesn't have Fira installed and I was using a locally downloaded html file. I am looking forward to seeing it with Fira.

> Screenshot 2024-10-26 at 6 41 24 PM
> Do you think it makes sense to change the finger point from an image (https://docs.racket-lang.org/reference/finger.png) to a unicode text, so that the color changes accordingly in the dark mode?
> 

This does make sense, I'll have to dig around in the repo to do this

> Have you checked documentation pages that display graphics (e.g., the output is a filled black square)? Do they work well?
> 

I have not. I can try out something. Do you have a good candidate page in mind?

> Do you plan to add an explicit button to switch the theme (probably not as a part of this PR, but in the future)?

I hadn't planned this far ahead. This just uses a media query which goes off of the operating system settings. While a button to toggle would be nice, my hunch is that it may require more scribble internal knowledge than I currently have.

soegaard:
It's a nice start, but I think it will be quite an amount of work to get all pages to look reasonable with the new color scheme.

On the screen shot I noticed the color of the hand hasn't changed from black to white:

image

image

When it comes to svg-images @sschwarzer and I looked at supporting dark mode in the glossary.
If the images are simple black and white images, one can use CSS to swap the colors.

https://git.sr.ht/~sschwarzer/racket-glossary/tree/main/item/fix-svgs.rkt

dr-neptune:
@soegaard I agree that it will be difficult to make everything look good, but I think this is an overfitting problem (overfitting to the light theme) that might be overcome with people having a dark theme in mind when crafting their docs. The finger and most images I've come across are `png`s which are not amenable to changing colors with `css`. 

As an example, here is documentation on `pict` (which embeds each image as a `png` file):

image

re: using the `svg` selector:
Do you have a reference page I could use? Right now I'm hesitant to blanket apply a style to `svg`s as I'm not sure exactly what that might change. Will changing the media-query for `svg`s to have `color: white`, `background-color: black` also remove color from `svg`s that have color?

@sorawee Given that the finger is referenced from the main racket repo, I would have to change the image to a text character there. Do you think that it would be worth changing the finger since it is a racket docs specific thing or would it be better to keep downstream it's own thing?

soegaard:
> @soegaard I agree that it will be difficult to make everything look good, but I think this is an overfitting problem (overfitting to the light theme) that might be overcome with people having a dark theme in mind when crafting their docs. 

That suggests to me that an explicit button to switch themes is needed.

Wrt. to png vs svg. Use the `++convert` flag for Scribble to get svgs instead of pngs.

```
++convert <fmt>
     prefer image conversion to <fmt> (in given order)
      <fmt> as one of: ps pdf svg png gif
```

> re: using the svg selector:
Do you have a reference page I could use? Right now I'm hesitant to blanket apply a style to svgs as I'm not sure exactly what that might change. Will changing the media-query for svgs to have color: white, background-color: black also remove color from svgs that have color?

The `color` properties affect the color for points and lines that haven't elsewhere been given a color.
For transparent svg's there is no need to use the background color property.

I don't have a reference page to point you to. Maybe the Glossary - but I can't remember if Stefan actually enabled it, or just experimented with it.

soegaard:
Schwarzer reminded me of this thread on dark mode:

https://todo.sr.ht/~sschwarzer/racket-glossary/4

jackfirth:
Getting even the basic cases working (as this PR already does) would be hugely valuable in my opinion. How do we feel about merging this without figuring out the SVG issue and defer that to follow-up work?

As for an explicit toggle: that does seem useful, but not _strictly_ necessary for an MVP. The operating system and browser already have those toggles. There's also browser extensions that give you per-site toggles in the client side if people really want to disable it without affecting other sites. That seems like a good enough workaround before we implement our own toggle.

I'm really excited for this feature, if you couldn't tell. My insomniac coding habits have made this an issue for me for years.

otherjoel:
> that might be overcome with people having a dark theme in mind when crafting their docs

I just want to say I am very opposed to this notion in particular. Authors are the wrong people on which to place the burden for getting good results in various schemes and themes the system might offer. If people want to use dark mode that’s fine but Scribble needs to do 100% of the work to make documents accessible in that scenario.

sschwarzer:
> Authors are the wrong people on which to place the burden for getting good results in various schemes and themes the system might offer.

I especially agree with this statement. It's difficult enough to come up with the actual content and it's hard to motivate oneself to put in more effort to support a dark mode (or any other theme).

Honestly, I guess I only invested so much time in dark mode for the glossary because I used to use dark mode myself and wanted to support current dark mode users. But I don't think this should be _expected_ from authors. Besides, dark mode support can also be difficult to test. For example, I was surprised that I couldn't get a good dark mode experience with the DarkReader plugin, but it worked with the dark mode in the browser.

So my recommendation would be to strive for some automatic "best effort" support in Scribble and not require extra markup from documentation authors (unless it's trivial to add, maybe). But even then there's still the risk that some automatism makes it _more_ difficult to get a good result because the automatism interferes with dark mode handling in operating systems, browsers and plugins in unexpected ways.

dr-neptune:
If I understand correctly, the way forward seems to be to make dark mode a toggleable button? The image handling pieces may need to come later

soegaard:
> If I understand correctly, the way forward seems to be to make dark mode a toggleable button? The image handling pieces may need to come later.

Yes. 

Alternatively, one could also make dark-mode an opt-in for authors.
That is, an author that wants dark-mode could require, say, `scribble/dark-mode`.
This way, the author is involved and expected to check that both modes work.
Even with this solution, a toggle would be good thing.

PR #508, opened 2025-05-28
add 'bare-anchors style property

stevebyan:
Add a `'bare-anchors` style property that causes the html renderer to omit the "Link to here" hover links on titles and section headers. This just leaves the bare anchor tag, without the "Link to here" decoration. 

This change removes the need for CSS to hide the link and disclose it only during a mouser hover. A user can still obtain a link to the title or section from the Table of Contents sidebar.

This change facilitates the use of Scribble html output in contexts other than Racket documentation.

### DIFF
--- a/scribble-doc/scribblings/scribble/core.scrbl
+++ b/scribble-doc/scribblings/scribble/core.scrbl
@@ -488,6 +488,9 @@ The recognized @tech{style properties} are as follows:
        to skip the generation of an entry for the part's title in the
        document index.}
 
+ @item{@racket['bare-anchors] --- For HTML, omits the ``Link to here''
+   hidden hover link on title and section headers.}
+
  @item{@racket[document-version] structure --- A version number for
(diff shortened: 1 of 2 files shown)


mflatt:
As implemented here, the `bare-anchors?` parameter looks unnecessary. The `(part-style? d 'bare-anchors)` check could be moved to `render-part-content` where `(bare-anchors?)` is used, right?

However, I think it would make sense for `bare-anchors?` to apply to all subparts, instead of just the immediate part, and then it only needs to be set at the top of a document to disable anchors everywhere. In that case, the parameter is useful, but it would need to be parameterized as `[bare-anchors? (or (bare-anchors?) (part-style? d 'bare-anchors))]`.

Am I following correctly? Does automatically propagating `'bare-anchors` to subparts seem like a good idea?

(Minor: there should be a `history` note for this change with a bump to the `scribble-lib` version number.)

PR #509, opened 2025-06-01
replace use of html tt with span of class 'stt'

stevebyan:
HTML5 has deprecated the 'tt' tag. This commit replaces it with a span of class 'stt'. Note that the Scribble @tt function has already been fixed to use a span of class 'tt''. These two remaining uses are styling the nbsp separator between the section heading number and the section heading text. One use is in the tocsub link for the section, and the other is in the section heading itself.

### DIFF
--- a/scribble-lib/scribble/html-render.rkt
+++ b/scribble-lib/scribble/html-render.rkt
@@ -767,7 +767,7 @@
                                               ,@(format-number
                                                  (collected-info-number
                                                   (part-collected-info p ri))
-                                                 '((tt nbsp)))))
+                                                 '((span ([class "stt"]) nbsp)))))
                                       '(""))
                                 ,@(if (toc-element? p)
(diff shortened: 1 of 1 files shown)


sorawee:
@stevebyan For the future though, you might want to render pages that use the feature more prominently. Here, `tt` is only used on whitespaces, so it's difficult to see its effect. But I think this is straightforward enough, and safe to merge.

Also, it would be nice to also test the manual style (`scribble/manual`) in addition to the base style. The manual style is arguably more important than the base style, since it's what most people use and see.

stevebyan:
This PR did not affect the Scribble “@tt” function. It already uses CSS rather than the tt tag. It only modifies the hard-coded formatting of section headers and their TOC entries, which only uses the tt tag on white space.

stevebyan:
Thanks for merging it.

PR #510, opened 2025-06-03
replace SAuthorList span with div

stevebyan:
Each author is a paragraph. Block elements are not allowed within an inline element. Change the span to a div.

### DIFF
--- a/scribble-lib/scribble/html-render.rkt
+++ b/scribble-lib/scribble/html-render.rkt
@@ -1201,7 +1201,7 @@
                  (if (null? auths)
                      null
                      `((div ([class "SAuthorListBox"])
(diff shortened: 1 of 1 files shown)


stevebyan:
I'll post images in my next comment, but to examine this in the browser of your choice, compare
https://www.byan-roper.org/steve/drafts/scrbl-sauthorlist/GR-tuning-fork-oscillators.html
to 
https://www.byan-roper.org/steve/drafts/scrbl-stock/GR-tuning-fork-oscillators.html

stevebyan:
Safari with this change:
Safari with SAuthorList-div

Safari with stock Scribble:
Safari with stock Scribble

sorawee:
Thanks!

PR #512, opened 2025-06-03
Use h1 tag for the title instead of h2

stevebyan:
From https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements: 
> A page should generally have a single h1 element that describes the content of the page (similar to the document's title element). 

Changes the html-renderer to use h1, h2, h3, etc in place of h2, h3, h4. 

Also change scribble.css so that the h1, h2, h3, etc heading font-sizes match the old h2, h3, h4, etc font-sizes.

### DIFF
--- a/scribble-lib/scribble/scribble.css
+++ b/scribble-lib/scribble/scribble.css
@@ -500,7 +500,7 @@ ol ol ol ol { list-style-type: upper-alpha; }
   position: relative;
   float: right;
   left: 2em;
-  top: -2.5em;
+  top: -2.25em;
   height: 0em;
   width: 13em;
   margin: 0em -13em 0em 0em;


sorawee:
Curious: where did you get these 1.4em, 1.17em, etc constants? Are they consistent across browsers? (IIRC, browsers are free to set the initial styling arbitrarily?)

stevebyan:
The constants come from theWhatWG spec:
https://html.spec.whatwg.org/multipage/rendering.html#sections-and-headings

The heading font sizes are apparently standardized in modern browsers:
https://stackoverflow.com/questions/5410066/what-are-the-default-font-sizes-in-pixels-for-the-html-heading-tags-h1-h2/70720104#70720104

stevebyan:
OK, it's fixed.

sorawee:
Thanks!

PR #519, opened 2025-06-27
WIP: add option to output xexpr of html (new source branch)

stevebyan:
Output the body as an xexpr, and provide various metadata, including TOC links, in a struct rather than in html.

I've refactored html-render.rkt to optionally output the body of the document as an xexpr rather than as html, and provide document metadata including the TOC links in a struct rather than in html. My goal is to allow use of scribble's html in something like Greg Hendershott's Tadpole, and so I want to provide the document content in a form that can be applied to an xexpr template. I also want to provide the TOC links in a form that can be rendered by site templates using pure CSS to provide a responsive view of the navigation links.

I'm submitting this pull request to gather some feedback on my approach, before committing to doing more work to make this something actually mergeable into Scribble. To pull out the TOC as data rather than as html, I've duplicated a lot of code from the render-toc-view and render-onthispage-contents methods. A finished pull request would refactor render-toc-view and render-onthispage-contents to use the output of my list-of-toc-view and list-of-onthispage-contents. I'm also not sure at this point how I want to represent the TOC contents. Using structs would probably be better than the current hacked nested lists.

Is this a feature the Racket developers would consider including in Scribble? Except for the duplicated code, does my implementation approach look reasonable?

Supercedes PR #498 
I mistakenly created the PR from the master branch in my fork. Now that a number of commits have been made to racket/scribble master, I need to reset my master back, pull the new commits, and merge in the new commits to my pull request.

### DIFF
--- a/scribble-lib/scribble/base-render.rkt
+++ b/scribble-lib/scribble/base-render.rkt
@@ -57,7 +57,8 @@
                 [extra-files null]
                 [image-preferences null]
                 [helper-file-prefix #f]
-                [keep-existing-helper-files? #f])
+                [keep-existing-helper-files? #f]
+                [xexpr-out? #f])
 
     (define/public (current-render-mode)
       '())
--- a/scribble-lib/scribble/html-render.rkt
+++ b/scribble-lib/scribble/html-render.rkt
@@ -1088,7 +1088,7 @@
           (define script-file-path
             (or (lookup-path script-file alt-paths) 
                 (install-file/as-url script-file)))
-          (unless  xexpr-out?
+          (unless xexpr-out?
             (if (bytes? prefix-file)
               (display prefix-file)
               (call-with-input-file*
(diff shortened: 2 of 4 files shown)


LiberalArtist:
FYI, you can now change this:
https://github.com/racket/scribble/blob/60787f79ac718e00a98d784d7b8029ba760a1735/scribble-lib/info.rkt#L6
to something like:
```racket
["base" #:version "8.17.0.5"] ; for serializable X-expressions
```
to tell the package system you need a version including the serializable-X-expressions change.

Eventually, you will also want to increment the package version here:

https://github.com/racket/scribble/blob/60787f79ac718e00a98d784d7b8029ba760a1735/scribble-lib/info.rkt#L26

and add `@history[]` notes to the documentation so that other people can depend on your new feature.

stevebyan:
@LiberalArtist, thank you

PR #520, opened 2025-06-28
Enclose scribble parts in an html section tag

stevebyan:
Enhance the Scribble html renderer to enclose the `@title`, `@section`, `@subsection`, etc parts in an html <section> tag. This enables better support for assistive technologies, and some CSS layout strategies such as a "block-based" design. For the latter, see Shriram Krishnamurthi's thread on the Racket Discord.

The section tag `class` attribute is set to the level of the part in the of `SsectionLevel<n>`, and the `id` attribute is set to `section <the section number>`. The `id` attribute is necessary for "block-based" design.

### DIFF
--- a/scribble-lib/scribble/html-render.rkt
+++ b/scribble-lib/scribble/html-render.rkt
@@ -1128,7 +1128,12 @@
             [else l]))
         (let ([number (collected-info-number (part-collected-info d ri))])
           (add-title-and-content-wrapper
-           `(,@(let ([pres (extract-pretitle d)])
+           `((section ([class ,(format "SsectionLevel~a" (number-depth number))]
+                       [id ,(format "section ~a"
+                                    (if (null? number)
+                                        "0"
+                                        (car (format-number number ""))))])
+             ,@(let ([pres (extract-pretitle d)])
                  (append-map (lambda (pre)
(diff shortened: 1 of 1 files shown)


stevebyan:
Oops, looks like I have some more work to do, some tests are failing.

mflatt:
Looking at the implementation and then back at the rendered files, I see that I was wrong: https://docs.racket-lang.org/reference/eval-model.html uses `h3`, not `h2`. Probably that's something else we'd want to fix in the long run!

Meanwhile, it doesn't look like the right information is available. I think `part-nesting-depth` turns out to be more about subdirectories and not nesting of parts within parts. Maybe the call to `render-part` in `render-one-part` needs to be `parameterize`d to communicate the depth of the enclosing part.

Given the current state, I'd be ok with merging the PR as-is and leaving relative depth to a future option. But I'm also happy if you think it's better to add a little more now to make the class name a relative depth.

stevebyan:
> Looking at the implementation and then back at the rendered files, I see that I was wrong: https://docs.racket-lang.org/reference/eval-model.html uses `h3`, not `h2`. Probably that's something else we'd want to fix in the long run!

Yeah, it's the title of the page, it should be `h1`.

mflatt:
Thanks!

PR #521, opened 2025-06-28
Declare deprecated aliases

jackfirth:


jackfirth:
Hmm. Is there a way for `define-deprecated-alias` to raise an error if the target is unbound in the surrounding module or definition context _and_ not require the target be bound before the `define-deprecated-alias` form? That would be the friendliest API for users.

jackfirth:
Is there any way to achieve the same thing in an internal definition context?

mflatt:
Do you mean an internal definition context that's in a module? Using `syntax-local-lift-module-end-declaration` should do the right thing when lifting a reference that happens to be in a definition context, as long as that definition context is also in a module as illustrated below. But maybe I don't understand what you're after.

```
#lang racket

(define-syntax (check-later stx)
  (syntax-case stx ()
    [(_ id)
     (when (syntax-transforming-module-expression?)
       (syntax-local-lift-module-end-declaration #'(check-now id)))
     #'(void)]))

(define-syntax (check-now stx)
  (syntax-case stx ()
    [(_ id)
     (unless (identifier-binding #'id)
       (raise-syntax-error #f "not good" #'id))
     #'(void)]))

(let ([ok "ok"])
  (check-later ok))

(let ()
  (check-later ok)
  (define ok "ok")
  (void))

(let ()
  (define something-else 0)
  (check-later ok-too))

(define ok-too 10)
```

PR #556, opened 2026-04-08
Error using #:auto-value of number

bdeket:
Make sure that `syntax-span` is not false before checking length

currently rendering scribble with an auto-value (or using #:value with other syntax objects that are numbers) fails.
This fixes the direct problem for me, but I'm not sure it's the best approach.

Example:
```
#lang scribble/manual

@(module a racket/base
   (provide :a :b)
   (define :a 4.8)
   (define :b 'symbol))

@(require (for-label (submod "." a)))

@defthing[:b symbol? #:auto-value]
@defthing[:a number? #:value 3.5]
@defthing[:a number? #:auto-value] @;; this fails
```
tested using "Welcome to DrRacket, version 9.1.0.11 [cs]." on windows

### DIFF
--- a/scribble-lib/scribble/racket.rkt
+++ b/scribble-lib/scribble/racket.rkt
@@ -280,8 +280,9 @@
          [(and (number? sc)
                (inexact? sc))
           (define s (iformat "~s" sc))
-          (if (= (string-length s)
-                 (- (syntax-span c) 2))
+          (if (let ([span (syntax-span c)])
+                (and span (= (string-length s)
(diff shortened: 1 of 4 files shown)


bdeket:
Hi,
could someone have a look at this?
Am I using `#:auto-value` wrong, or should I be adding tests? ( if so, where? )
kr. B.

mflatt:
Thanks! I added to the test document, and now I'll merge.

bdeket:
Thank you.

PR #565, opened 2026-07-09
Fix hyperref label mismatch for tech/def cross-references in LaTeX

wilbowma:
When rendering `@tech` links with `#:tag-prefixes`, the resolved destination tag (e.g. `(tech "chp1:" "book:" "widget")`) differs from the raw link-element tag (e.g. `(tech ("book:" "chp1:" "widget"))`), causing `\hyperref` to reference a label that does not exist.

The section/part rendering path appears to handle this by using `(vector-ref dest 1)` from the resolved destination rather than `(link-element-tag e)`. Adapt the same pattern for tech/def refs.


mflatt:
Thanks!

PR #566, opened 2026-07-13
add typst-watch to monitor changes to scribble files

capfredf:
We added a small even loop to monitor scribble file changes.
We used  `dynamic-rerequire` to reload related doc modules

### DIFF
--- a/scribble-lib/scribble/run.rkt
+++ b/scribble-lib/scribble/run.rkt
@@ -2,6 +2,8 @@
 (require "xref.rkt"
          "render.rkt"
          racket/cmdline
+         racket/system
+         racket/rerequire
          raco/command-name
          (prefix-in text:     "text-render.rkt")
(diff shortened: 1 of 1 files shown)


capfredf:
@mflatt Do you think I should move the code to where the typst renderer is located? 

Another related question: should I put test code in  `tests/scribble/typst.rkt`? It seems that using a command to test this feature in that file does not fit.

mflatt:
Would it make sense to have a general `--watch` flag instead of having a Typst-specific flag? It seems like watching could work in general.

Maybe it would also make sense to detect changes to dependencies of the main document file. The `show-import-tree` from `syntax/moddep` might be good for building the set of dependencies. Or this might create more problems than it solves; with dependencies, then there are more concerns with compiling modules, and maybe it's not ok to assume that every state is well-formed and should be loaded/rendered immediately.

For a test, I agree that it would be better to have a separate test module.

PR #568, opened 2026-08-07
Render lsquo as a backtick in content->string

slbls:
`content->string` renders `lsquo` as the literal text `lsquo`. The HTML renderer uses `content->string` for document titles, which causes [the Racket Guide’s quasiquoting page](https://docs.racket-lang.org/guide/qq.html) to have `lsquo` in its title.

Map `lsquo` to a backtick, matching the existing `rsquo` mapping to an apostrophe. Add a regression test through `decode-string`.

Tests

- `raco setup --check-pkg-deps --pkgs scribble-text-lib scribble-html-lib scribble-lib scribble-doc scribble-test`
- `raco test --drdr --package scribble-text-lib scribble-html-lib scribble-lib scribble-doc scribble-test`

### DIFF
--- a/scribble-lib/scribble/core.rkt
+++ b/scribble-lib/scribble/core.rkt
@@ -698,6 +698,7 @@
                         [(mdash) "---"]
                         [(ndash) "--"]
                         [(ldquo rdquo) "\""]
+                        [(lsquo) "`"]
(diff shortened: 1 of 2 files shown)


mflatt:
Thanks!

PR #569, opened 2026-08-08
Render emphasis, subscripts, and superscripts as semantic HTML

slbls:
While adapting The Racket Guide to EPUB 3 for myself locally (reading using Apple Books), I noticed that Scribble renders `@emph`, `@superscript`, and `@subscript` as styled `<span>` elements.

Render them as `<em>`, `<sup>`, and `<sub>`. Keep `class="emph"` and the current subscript and superscript presentation. Add `line-height: inherit` to preserve the existing line-box behavior. Explicit `alt-tag` values, links, and custom string styles retain their current behavior.

This follows [PR #220](https://github.com/racket/scribble/pull/220), which made `@emph` logical and nestable. Its discussion proposed `<em>` output and [identified the HTML renderer as the implementation point](https://github.com/racket/scribble/pull/220#issuecomment-553451731). That PR already added Markdown support.

### DIFF
--- a/scribble-lib/scribble/html-render.rkt
+++ b/scribble-lib/scribble/html-render.rkt
@@ -1672,6 +1672,13 @@
               (let ([s (content-style e)])
                 (and (style? s)
                      (style->tag s)))]
+             [semantic-tag
+              (and (symbol? name)
+                   (case name
+                     [(emph) 'em]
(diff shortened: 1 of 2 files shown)


mflatt:
Thanks!

PR #573, opened 2026-09-13
autobib: include "." after doi, to clearly separate from notes.

fare:
- autobib.rkt: include "." after doi, to clearly separate from notes.
- autobib.tex: use \Needspace instead of \parbox, much improving support for long "note" fields.

This much improves the output for the heavily annotated bibliography of my book LtUO.


fare:
PS: I'm tempted to have a clanker improve the parsing of bibtex to support (a) emphasis, in notes, (b) {} in titles to prevent lowercasification when using LaTeX.

Also, maybe to improve autobib so multi-paragraph notes are a thing, and maybe crossing page boundaries within a bibliography entry.

These two would also much improve LtUO.

_PPS_: I'm still thinking about (a), but at least the basics of (b) was apparently done by #571.

fare:
Weird: the CI failed on Racket 9.3.0.8, but at home it just worked on v9.3.0.8-2026-09-14-06fd59e02c [cs].

_PS_: I had forgotten to update a test.

mflatt:
The change to add `.` after a DOI would cause papers written with Scribble+autobib to render differently than Latex+bibtex (in the "acmart" style). Would it work to add `.` only when there's a note? Or maybe it should be configurable?

fare:
Yes, adding a "." only if there's a note makes a lot sense, especially for backwards compatibility. Thanks for the suggestion.

_PS_: also rebasing to the latest master.

